PR #1, opened 2018-12-23
Adding MIT LICENSE file

weaverryan:
This repository is a sub-tree split from symfony/symfony-docs originally, which is licensed under a Creative Commons Attribution-Share Alike 3.0 Unported License. However, all of that code has been deleted, and the sub-tree split was only to maintain the new history of this library.

The following authors have code that exists in this repository and we need their approval.

```
git ls-files | while read f; do git blame --line-porcelain $f | grep '^author '; done | sort -f | uniq -ic | sort -n
```

* [X] @weaverryan 
* [ ] @nikophil

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+The MIT license
+
+Copyright (c) 2018-present Ryan Weaver
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


weaverryan:
I, @weaverryan, approve changing the this library to MIT effective immediately.

nikophil:
I, @nikophil , approve changing the this library to MIT effective immediately.

PR #2, opened 2018-12-27
test external url status code

nikophil:
this is actually some draft / POC for ci check external urls
related issue: https://github.com/weaverryan/symfony-docs/issues/17

### DIFF
--- a/src/Command/CheckUrlsCommand.php
+++ b/src/Command/CheckUrlsCommand.php
@@ -1,0 +1,157 @@
+<?php declare(strict_types=1);
+
+namespace SymfonyDocs\Command;
+
+use Doctrine\RST\Builder;
+use Doctrine\RST\Event\PostNodeRenderEvent;
+use Doctrine\RST\Event\PostParseDocumentEvent;
+use Doctrine\RST\Event\PreBuildRenderEvent;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+use SymfonyDocs\CI\UrlChecker;
+use SymfonyDocs\KernelFactory;
+
+class CheckUrlsCommand extends Command
+{
+    protected static $defaultName = 'build:check-urls';
+
+    /** @var SymfonyStyle */
+    private $io;
+    private $filesystem;
+    private $finder;
+    /** @var ProgressBar */
+    private $progressBar;
+    /** @var Builder */
+    private $builder;
+    /** @var OutputInterface */
+    private $output;
+    private $sourceDir;
+    private $htmlOutputDir;
+    private $parsedFiles = [];
+    private $urlChecker;
+
+    public function __construct()
+    {
+        parent::__construct(self::$defaultName);
+
+        $this->filesystem = new Filesystem();
+        $this->finder     = new Finder();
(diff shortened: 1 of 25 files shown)


OskarStark:
Hmm this could bring up some false positives and should only be checked once and cache the result. Otherwise it could break builds where the change is unrelated 🤔

src/Command/CheckUrlsCommand.php
```diff
@@ -0,0 +1,157 @@
...
+        $this->urlChecker = new UrlChecker();
+    }
+
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->addArgument('source-dir', InputArgument::OPTIONAL, 'RST files Source directory', getcwd())
+            ->addOption('parse-only', null, InputOption::VALUE_OPTIONAL, 'Parse only given directory for PDF (directory relative from source-dir)', null);
```
weaverryan: I think this is unused

src/Command/CheckUrlsCommand.php
```diff
@@ -0,0 +1,157 @@
...
+            $this
+        );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->finder->in($input->getArgument('source-dir'))
+            ->exclude(['_build', '.github', '.platform', '_images'])
+            ->notName('*.rst.inc')
+            ->name('*.rst');
```
weaverryan: Yep, I see what you mean about needing some DRY. This is begging to be refactored into a service and re-used :)

src/Renderers/SpanNodeRenderer.php
```diff
@@ -32,7 +37,16 @@ public function link(?string $url, string $title, array $attributes = []): strin
...
+            $attributes['class'] = sprintf('reference %s', $this->isExternalUrl($url) ? 'external' : 'internal');
+        }
+
+        if (
+            $this->urlChecker &&
+            $this->isExternalUrl($url) &&
+            false === strpos($url, 'http://localhost') &&
+            false === strpos($url, 'http://192.168')
+        ) {
+            $this->urlChecker->checkUrl($url);
```
weaverryan: We're overriding & extending this class because there is no clean way to do it via an event, right (like you described in your comment)? If so, we should create an issue on rst-parser about that - Jon will not like this. I mean, he doesn't want us to need to override classes like this to get our job done - I think he'll want to expand the extension point.

src/Renderers/SpanNodeRenderer.php
```diff
@@ -44,4 +58,13 @@ public function link(?string $url, string $title, array $attributes = []): strin
             ]
         );
     }
+
+    public function isExternalUrl($url): bool
+    {
+        if (0 === strpos($url, 'http://') || 0 === strpos($url, 'https://')) {
```
weaverryan: Just look for `://`? And simplify:

```php
return strpos($url, '://') !== false;
```

src/Renderers/SpanNodeRenderer.php
nikophil: yes, you're absolutely right :)

nikophil: i'm gone make a PR for this

nikophil: https://github.com/doctrine/rst-parser/pull/89

nikophil: once this will be accepted, we could get rid of this object and use the new event

src/Command/CheckUrlsCommand.php
nikophil: ok, ive made a BIG drying and cleanup tour...
maybe to much...  :smile:  we'll see when we'll create new commands

src/Command/CheckUrlsCommand.php
nikophil: done

src/Renderers/SpanNodeRenderer.php
nikophil: done

PR #3, opened 2018-12-27
Export for pdf

nikophil:
Here is some more or less advanced work for exporting things for pdf
are you willing that i provide some functional test of this ?
related issue: https://github.com/weaverryan/docs-builder/issues/9


src/Command/BuildDocsCommand.php
```diff
@@ -129,7 +129,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         foreach ($this->finder as $file) {
             $htmlFile = str_replace([$this->sourceDir, '.rst'], [$this->htmlOutputDir, '.html'], $file->getRealPath());
-            if (!$this->filesystem->exists($htmlFile)) {
+
+            $firstLine = fgets(fopen($file->getRealPath(), 'r'));
+
+            if (!$this->filesystem->exists($htmlFile) && ':orphan:' !== trim($firstLine)) {
```
weaverryan: When we refactor the logic of this command into a service, we could update our functional test to check for this - i.e. have 2 files that aren't rendred, but make 1 an orphan, and check that we correctly see 1 unrendered file.

Also, this is *exactly* the kind of stuff that we should not need to do :). We should be able to get the "meta" for this document and check if it is an "orphan". If we're missing that from the meta (I think you said this was the case), we should open an issue on the rst-parser repo.

src/Command/BuildDocsCommand.php
```diff
@@ -129,7 +129,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         foreach ($this->finder as $file) {
             $htmlFile = str_replace([$this->sourceDir, '.rst'], [$this->htmlOutputDir, '.html'], $file->getRealPath());
-            if (!$this->filesystem->exists($htmlFile)) {
```
weaverryan: It's not part of this command, but on general comment: I think this command should have some `--render-single-file=full.html` command. IF this is set, then, in addition to the normal rendering, we would also render a single `full.html` file (we would normally only use this with the `--parse-only` option while rendering one of the PDF's, but I don't think that should be a requirement).

src/Generator/HtmlForPdfGenerator.php
```diff
@@ -45,8 +46,99 @@ public function generateHtmlForPdf(string $htmlDir, string $parseOnly/*, Progres
...
+                if (0 === strpos($matches[2], 'http')) {
+                    return sprintf('<em><a href="%s">%s</a></em>', $matches[2], $matches[2]);
+                }
+
+                return sprintf('<em>%s</em><span class="footnote"><code>%s</code></span>', $matches[2], $matches[1]);
+            },
+            $content
+        );
+
+        return $content;
```
weaverryan: I would say that, instead of a functional test, this needs a good old-fashioned unit test :). Well, it will probably feel a bit like a functional test - "given this raw HTML, here is the cleaned up HTML" - but yes, 100% - this stuff needs a test.

src/Generator/JsonGenerator.php
```diff
@@ -91,6 +91,16 @@ private function generateToc(MetaEntry $metaEntry, ?array $titles): array
 
     private function guessNext(string $parserFilename): ?array
     {
+        $meta       = $this->getMeta($parserFilename);
+        $parentFile = $meta->getParent();
+
+        if ('index' === $parentFile && \count($tocs = $meta->getTocs()) === 1 && \count($tocs[0]) > 0) {
```
weaverryan: Can you explain this with a comment? This is the type of stuff that I definitely want a test covering - to help us know why this is here and not accidentally break it :)

src/Command/BuildDocsCommand.php
nikophil: https://github.com/doctrine/rst-parser/issues/88
;)

src/Command/BuildDocsCommand.php
nikophil: btw i don't like much `parse-only` name ;)

weaverryan: Yea, me either... ``--parse-path`? "only" makes it sound like it's a boolean - everytime I see `string $parseOnly` I get confused :)

src/Generator/HtmlForPdfGenerator.php
```diff
@@ -2,41 +2,141 @@
...
-            throw new \InvalidArgumentException('File "%s" does not exist', $indexFile);
+            throw new \InvalidArgumentException(sprintf('File "%s" does not exist', $indexFile));
         }
 
+        // extracting all files from index's TOC, in the right order
         $parserFilename = $this->getParserFilename($indexFile, $buildContext->getHtmlOutputDir());
         $meta           = $this->getMeta($environments, $parserFilename);
-        dump(current($meta->getTocs()));
+        $files          = current($meta->getTocs());
+        array_unshift($files, sprintf('%s/index', $buildContext->getParseSubPath()));
```
weaverryan: At this point `$meta->getTocs()` would contain ALL the Tocs right? Including things not in the sub-path? 

And then we prepend `$files` with the parse-path index file? I'm missing something here...

src/Generator/HtmlForPdfGenerator.php
```diff
@@ -2,41 +2,141 @@
...
+            $meta = $this->getMeta($environments, $file);
+
+            $filename = sprintf('%s/%s.html', $htmlDir, $file);
+            if (!$fs->exists($filename)) {
+                throw new \LogicException(sprintf('File "%s" does not exist', $filename));
+            }
+
+            // extract <body> content
+            $crawler     = new Crawler(file_get_contents($filename));
+            $fileContent = $crawler->filter('body')->html();
```
weaverryan: A shame there's no way to fetch the built-content directory from `rst-parser`...

src/Generator/JsonGenerator.php
weaverryan: longer-term, I'd like to think about some other way to do this. Like maybe these "index" pages... that are not *really* part of the docs - are given some sort of option on top `:index-page:` - not sure... I just don't love enforcing this convention deep down here.

src/Generator/HtmlForPdfGenerator.php
nikophil: not ALL... only the toc from the `parseSubPath`

Here we're grabbing the index.html from the subPath
```php
$basePath  = sprintf('%s/%s', $buildContext->getHtmlOutputDir(), $buildContext->getParseSubPath());
$indexFile = sprintf('%s/%s', $basePath, 'index.html');
```

Then, here we're grabbing all files from this index
```php
$parserFilename = $this->getParserFilename($indexFile, $buildContext->getHtmlOutputDir());
$meta           = $this->getMeta($environments, $parserFilename);
$files          = current($meta->getTocs());
```

And because we want the index file to be included, we're prepending it to the array:
```php
array_unshift($files, sprintf('%s/index', $buildContext->getParseSubPath()));
```

src/Generator/HtmlForPdfGenerator.php
nikophil: actually because we're working from the already-generated html files, that seems impossible...
Or it would expose some services which does that work... but the `DomCrawler` already exists ;)

weaverryan: Yea, like if rst-parser "stored" the final HTML somewhere (Meta is probably not the right place, but you get the idea) - and we could "query" for it from the Builder after - just something to think about for later.

nikophil: this was my first thought when i've started to do this feature... but html is never stored in any vars...

PR #4, opened 2018-12-27
CI check - code sniffer

nikophil:
here is a draft / POC for using phpcs in the builder
see https://github.com/weaverryan/docs-builder/issues/7


OskarStark:
I would recommend to use Php-CS-Fixer with the Symfony config or why do you propose code sniffer?

composer.json
```diff
@@ -6,6 +6,9 @@
     "autoload": {
         "psr-4": {
             "SymfonyDocs\\": "src"
+        },
+        "psr-0": {
+            "PHP_CodeSniffer\\": "vendor/squizlabs/php_codesniffer"
```
OskarStark: I guess this can be removed

nikophil:
Yep you're right, and that's what @weaverryan told me as well.
I'm afraid i just didn't know the existence of this one (which has a real better integration to symfony)

i've modified my code - and created a listener to use it.
I tried all possible events offered by the parser but none of them is satisfying.
If you check the listener's code, i've put some annotation above each one.

I think the best approach would be to ask Jon if it's possible to inject the `Environment` inside the nodes, in order to now where is coming each node. WDYT ?

composer.json
nikophil: :+1:

weaverryan:
Yea, very good experimentation here. Yes, we need access to the Environment. I left a comment on your issue: https://github.com/doctrine/rst-parser/issues/84#issuecomment-451338457

javiereguiluz:
Closing because we already have a source code validator in place. See https://github.com/symfony/symfony-docs/blob/61d8041981f96a7aea910db3b335b1c8d62aaf5e/.github/workflows/ci.yaml#L80

In any case, thanks for this and all your other contributions to this project 🙌

PR #15, opened 2019-01-04
Dependency injection

### LINKED ISSUE #11: Change namespace
namespace should be changed from `SymfonyDocs` to `SymfonyDocsBuilder`

### LINKED ISSUE #12: Use symfony's DI
maybe it's time to use symfony's DI, let's try it :)

### LINKED ISSUE #14: use env var SYMFONY_VERSION
the env var  `SYMFONY_VERSION` should be used to handle which version of the doc might be parsed

nikophil:
Lots of thing here:
- use of dependency injection (fixes #12)
- use of SYMFONY_VERSION  env var (fixes #14)
- namespace changed (fixes #11)
- i've created some kind of `ParameterBag` that will hold all the options we're passing through the command, and that we need in every other services - this permits us to use depency injection everywhere
- get rid of `Configuration` service, in favor of container's parameter bag
- fixes tests

@weaverryan that would be nice to discuss around this PR which changes a lot of things and that needs to be merged before all others.
thanks :)


src/Command/BuildDocsCommand.php
```diff
@@ -39,12 +40,29 @@ class BuildDocsCommand extends Command
...
-    public function __construct()
-    {
+    /** @var JsonGenerator */
+    private $jsonGenerator;
+    /** @var HtmlForPdfGenerator */
+    private $htmlForPdfGenerator;
+    /** @var ParameterBag */
+    private $parameterBag;
+    /** @var KernelFactory */
+    private $kernelFactory;
```
weaverryan: No need for the phpdoc here - as it’s inplied by the required constructor args

src/Application.php
```diff
@@ -0,0 +1,64 @@
...
+use SymfonyDocsBuilder\Command\BuildDocsCommand;
+
+class Application
+{
+    /** @var BaseApplication */
+    private $application;
+    /** @var BuildDocsCommand */
+    private $buildDocsCommand;
+
+    public function __construct(BaseApplication $application, BuildDocsCommand $buildDocsCommand)
```
weaverryan: We should pass an array of commands here - as we’ll have many. I’m fine if we need to manually register these as an argument in the services file - being explicit (instead of adding more magic layers like a compiler pass) has big advantages of simplicity when working on a lib.

src/Application.php
```diff
@@ -0,0 +1,64 @@
...
+            'The symfony version of the doc to parse.',
+            false === getenv('SYMFONY_VERSION') ? '4.1' : getenv('SYMFONY_VERSION')
+        );
+        $this->application->getDefinition()->addOption($inputOption);
+        $this->application->add($this->buildDocsCommand);
+
+        return $this->application->run($input);
+    }
+
+    public static function getContainer(string $version): ContainerBuilder
```
weaverryan: createContainer()

src/ParameterBag.php
```diff
@@ -0,0 +1,240 @@
...
+ */
+class ParameterBag implements \IteratorAggregate, \Countable
+{
+    /**
+     * Parameter storage.
+     */
+    protected $parameters;
+
+    /** @var bool */
+    private $initialized = false;
```
weaverryan: Needed? It looks like everything is required (and it should be). I do see a call to initialize() in a different class - I don’t like that - if this is a service that holds some variables, it should be fully populated immediately by the container. Also... if we can think of a better name that describes what it holds in our app, I’d like that too :)

src/ParameterBag.php
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * ParameterBag is a container for key/value pairs.
+ */
+class ParameterBag implements \IteratorAggregate, \Countable
+{
+    /**
+     * Parameter storage.
+     */
+    protected $parameters;
```
weaverryan: private

src/ParameterBag.php
nikophil: the point here is that we can't populate it immediately by the container, because the parameters it holds are the ones we're passing to the command as parameter (sourceDir, outputDir...)
these are parsed once the command is started (and the container's parameterBag is frozen...)

nikophil:
> Some comments to get us started! One important thing - and I think you’ve done it - is to make sure that the code could still be used directly without the container or command application. For example, when this lib is required as a dependency, we should be able to use the parser to parse a single file without too much trouble.

i'm not afraid about that, everything could be ran without the help of the DI

conf.json
```diff
@@ -1,6 +1,6 @@
 {
     "version": "4.0",
```
weaverryan: Is this key still used?

conf.json
```diff
@@ -1,6 +1,6 @@
...
-    "symfony_api_url": "https://api.symfony.com",
+    "symfony_api_url": "https://api.symfony.com/%s",
     "php_doc_url": "https://secure.php.net/manual/en",
     "symfony_doc_url": "https://symfony.com/doc/%s"
```
weaverryan: In general, this all kinda looks like config that can be hardcoded - these URLs would never change, right?

src/Generator/HtmlForPdfGenerator.php
```diff
@@ -14,39 +13,30 @@ class HtmlForPdfGenerator
...
-    public function __construct(array $documents)
-    {
-        $this->environments = array_map(
-            function (DocumentNode $document) {
-                return $document->getEnvironment();
-            },
-            $documents
-        );
-    }
+    public function generateHtmlForPdf(
```
weaverryan: When array is the type-hint, we should add an `@param` phpdoc with what it actually is - e.g. `Document[]`

Issue #10, opened 2019-01-02
Ability to do cached/smarter builds


nikophil:
see https://github.com/weaverryan/symfony-docs/issues/18

nikophil:
discussion opened on the parser repo:
https://github.com/doctrine/rst-parser/issues/85

weaverryan:
This is now merged upstream :)

weaverryan:
Oh, actually, small thing, we need a flag to be able to build but ignore the cache. There is already a method on Configuration to control this

javiereguiluz:
Closing as fixed because we have this method to disable the cache: https://github.com/symfony-tools/docs-builder/blob/f34eaabf76580e54397ac1bf5b1fa2c3053d6a77/src/BuildConfig.php#L91

PR #22, opened 2019-02-03
Fix cache issue

### LINKED ISSUE #21: json generator bug if using cache
JSON generation process is not smart enough to realize that it doesn't need to rebuild everything.

It should be as simple as: if you can't find an Environment for a file (or maybe there is a better way to know if it was parsed, like using the `ParseQueue`, which has the list of files that needed to be parsed), then skip - the JSON should still be there and should still be up to date (I suppose we could throw an error if we find that the JSON file is missing for some reason)

nikophil:
fixes #21 & #10


src/BuildContext.php
```diff
@@ -30,12 +31,13 @@ public function __construct(
         $this->symfonyDocUrl  = $symfonyDocUrl;
     }
 
-    public function initializeRuntimeConfig(string $sourceDir, string $htmlOutputDir, ?string $jsonOutputDir = null, ?string $parseSubPath = null)
+    public function initializeRuntimeConfig(string $sourceDir, string $htmlOutputDir, ?string $jsonOutputDir = null, ?string $parseSubPath = null, ?bool $disableCache = false)
```
weaverryan: `?bool` can just be `bool` i think

src/Generator/GeneratorTrait.php
```diff
@@ -44,4 +81,22 @@ private function getMeta(array $environments, string $parserFilename): MetaEntry
...
+    /**
+     * @param Environment[] $environments
+     */
+    private function useCacheForFile(string $parserFilename): bool
+    {
+        if ($this->buildContext->getDisableCache()) {
+            return false;
+        }
+
+        // if an environment exits for the given file, it means that the file has been built from the parser
```
weaverryan: exists

And we can:

```php
return !isset($this->environments[$parserFilename]);
```

src/Generator/GeneratorTrait.php
```diff
@@ -44,4 +81,22 @@ private function getMeta(array $environments, string $parserFilename): MetaEntry
 
         return $allMetas[$parserFilename];
     }
+
+    /**
+     * @param Environment[] $environments
+     */
```
weaverryan: `@param` is out of date

src/Generator/JsonGenerator.php
```diff
@@ -17,42 +18,55 @@ class JsonGenerator
...
-    ) {
-        $environments = $this->extractEnvironments($documents);
+    public function generateJson(array $documents, ProgressBar $progressBar)
+    {
+        $this->extractEnvironmentsAndCachedMetas($documents);
+
+        if (!$this->buildContext->getDisableCache()) {
+            $cachedMetasLoader = new CachedMetasLoader();
+            $cachedMetasLoader->loadCachedMetaEntries($this->buildContext->getHtmlOutputDir(), $metas = new Metas());
+        }
```
weaverryan: Is this needed now? Can't we use `$this->metas` after calling `extractEnvironmentsAndCachedMetas()`?

src/Generator/GeneratorTrait.php
nikophil: even if we expect an array of objects ?
this is usefull for phpstorm to do some completion...

src/Generator/JsonGenerator.php
nikophil: yes that's a mistake, sorry

nikophil: fixed

src/Generator/GeneratorTrait.php
nikophil: fixed

src/BuildContext.php
nikophil: done

PR #25, opened 2019-09-18
GitHub auto release

nikophil:


src/Command/GithubReleaseCommand.php
```diff
@@ -0,0 +1,162 @@
...
+        $this->addArgument('name', InputArgument::OPTIONAL, 'Release name', 'Symfony docs builder %s');
+        $this->addArgument('description', InputArgument::OPTIONAL, 'Symfony docs builder %s');
+    }
+
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $dotenv = new Dotenv();
+        $dotenv->load(__DIR__.'/../../.env');
+
+        if (empty($_ENV['GITHUB_API_TOKEN'])) {
```
weaverryan: For some edge-case reason, I'm told we should use `$_SERVER` for env vars :)

src/Command/GithubReleaseCommand.php
```diff
@@ -0,0 +1,162 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\HttpClient\Exception\ClientException;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\Process;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class GithubReleaseCommand extends Command
```
weaverryan: Having a command to release a command-line library feels weird to me - there's no reason, for example, that this should even been accessible when you're working with the phar file. Could we make this a flat `bin/create_release` file? We would still require the autoloader... and then probably we could refactor some of the code into "service classes" that are used in that script (so that the script is ultimately tiny). We could then *not* include this script nor those supporting classes in the phar file.

src/Command/GithubReleaseCommand.php
```diff
@@ -0,0 +1,162 @@
...
+
+        if (empty($_ENV['GITHUB_API_TOKEN'])) {
+            throw new RuntimeException('Please fill "GITHUB_API_TOKEN" in file "[PROJECT_DIR]/.env"');
+        }
+
+        $this->io     = new SymfonyStyle($input, $output);
+        $this->client = $this->createHttpClient($_ENV['GITHUB_API_TOKEN']);
+
+        $tag = $input->getArgument('tag');
+        if (!preg_match('/^v\d+\.\d+\.\d+$/', $tag)) {
```
weaverryan: +1 for a sanity check!

src/Command/GithubReleaseCommand.php
```diff
@@ -0,0 +1,162 @@
...
+
+            return (int) $response->toArray()['id'];
+        } catch (ClientException $exception) {
+            if (401 === $exception->getCode()) {
+                $message = 'Invalid token';
+            } else {
+                $message = 'Maybe the tag name already exists?';
+            }
+
+            throw new RuntimeException(sprintf('Error while trying to create release: %s.', $message), 0, $exception);
```
weaverryan: nice

src/Command/GithubReleaseCommand.php
```diff
@@ -0,0 +1,162 @@
...
+            $response = $this->client->request(
+                'POST',
+                sprintf('https://api.github.com/repos/%s/%s/releases', self::GITHUB_USER, self::GITHUB_REPO),
+                [
+                    'json' => [
+                        'tag_name'         => $this->tag,
+                        'target_commitish' => 'master',
+                        'name'             => $this->name,
+                        'description'      => $this->description,
+                        'draft'            => false,
```
weaverryan: Hmm. Could we set this `draft` => true, then upload the asset, then modify it to `draft=>false`? That's minor... but it would eliminate any time when the release is published and has *no* phar attached.

composer.json
```diff
@@ -19,7 +19,9 @@
         "symfony/dom-crawler": "^4.1",
         "symfony/css-selector": "^4.1",
         "symfony/console": "^4.1",
-        "twig/twig": "^2.7.3"
+        "twig/twig": "^2.7.3",
+        "symfony/http-client": "^4.3",
+        "symfony/dotenv": "^4.3"
```
weaverryan: move to require-dev

PR #26, opened 2020-10-26
Do not include file location in the assets url

wouterj:
When rendering locally, relative file paths also work fine (at least for me).


weaverryan:
Works for me, and we can revisit when we have a use-case

PR #27, opened 2021-02-22
Updated dependencies

javiereguiluz:


src/DocsKernel.php
```diff
@@ -36,8 +36,6 @@ public function initBuilder(Builder $builder): void
             $builder->getConfiguration()->getEventManager(),
             $builder->getErrorManager()
         );
-
-        $builder->setScannerFinder($this->buildContext->createFileFinder());
```
weaverryan, line 40: Is this not needed anymore? It was done before to take into account which files to ignore - https://github.com/weaverryan/docs-builder/blob/master/src/BuildContext.php#L136-L147

javiereguiluz: If I restore this line, tests show 6 errors like this one:

```
1) App\Tests\Command\BuildDocsCommandTest::testBuildDocsDefault
Error: Call to undefined method Doctrine\RST\Builder::setScannerFinder()

/Users/javier/work/docs-builder/src/DocsKernel.php:40
/Users/javier/work/docs-builder/vendor/doctrine/rst-parser/lib/Builder.php:76
/Users/javier/work/docs-builder/src/Command/BuildDocsCommand.php:138
/Users/javier/work/docs-builder/vendor/symfony/console/Command/Command.php:256
/Users/javier/work/docs-builder/vendor/symfony/console/Tester/CommandTester.php:76
/Users/javier/work/docs-builder/tests/Command/BuildDocsCommandTest.php:98
/Users/javier/work/docs-builder/tests/Command/BuildDocsCommandTest.php:37
```

Without this line, tests pass.

nikophil: this method only exist on `master` branch in `doctrine/rst-parser`

weaverryan: I'm trying to track down what's going on over here: https://github.com/doctrine/rst-parser/issues/132

weaverryan: Hopefully this will get merged, and then we'll be good https://github.com/doctrine/rst-parser/pull/133

In fact, MANY subtle features I added to rst-parser are not currently there in `0.2`. So after #133 is merged, we'll need to update to use `0.2` here and... some subtle things may definitely change.

greg0ire: I just released 0.3.0 with all the features you are missing, so that's what you'll want to update to.

javiereguiluz:
This PR is too out-of-sync. Closing in favor of #61.

PR #28, opened 2021-02-22
Replace Guzzle by Symfony HttpClient

javiereguiluz:
There's nothing wrong with Guzzle, but standardizing the code on Symfony components makes things easier to maintain.

### DIFF
--- a/src/CI/UrlChecker.php
+++ b/src/CI/UrlChecker.php
@@ -11,21 +11,21 @@ declare(strict_types=1);
 
 namespace SymfonyDocsBuilder\CI;
 
-use GuzzleHttp\Client;
-use GuzzleHttp\Exception\GuzzleException;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\Exception\HttpExceptionInterface;
 
 final class UrlChecker
 {
     private $invalidUrls = [];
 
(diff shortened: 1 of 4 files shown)


src/CI/UrlChecker.php
```diff
@@ -11,21 +11,20 @@
...
     {
-        $httpClient = new Client(['timeout' => 10]);
+        $httpClient = HttpClient::create(['timeout' => 10]);
 
         try {
-            $response = $httpClient->get($url, ['http_errors' => false]);
+            $response = $httpClient->request('GET', $url);
             $statusCode = $response->getStatusCode();
-        } catch (GuzzleException $exception) {
+        } catch (\RuntimeException $e) {
```
nikophil, line 27: shouldn't we narrow catch to `\Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface`?

javiereguiluz: I've changed it to `HttpExceptionInterface` ... if we catch only Transport exception we'd miss Server exception, Redirection exception, etc.

nikophil: indeed, i meant this one :+1:

nikophil (approved):
LGTM :+1:

OskarStark (approved):
Looks good 👍🏻😎

PR #30, opened 2021-02-22
Remove the PHAR-based releases

javiereguiluz:
Sorry for removing all this ... but for symfony.com (and I guess for other use-cases too) a normal distribution based on installable Composer packages is simpler. My guess is that this is better for this project too, because we remove some complexity and don't rely on third-party "moving parts" which can break, change or become unreliable. Thanks!


weaverryan:
I'm not convinced yet. I created this system to minimize the diff on symfony.com's codebase for integrating this. I'd prefer to (if possible) get symfony.com to use the docs-builder first. And then, if that implementation doesn't rely on the phar, we can remove it :).

PR #33, opened 2021-02-25
Use Symfony string for most string operations

javiereguiluz:
I think the improved readability is better than the slight performance decrease ... but I'm open to read different opinions. Thanks!

### DIFF
--- a/src/CI/MissingFilesChecker.php
+++ b/src/CI/MissingFilesChecker.php
@@ -33,7 +33,7 @@ class MissingFilesChecker
         $orphanedFiles = [];
 
         foreach ($finder as $file) {
-            $sourcePath = u($file->getPathname())->after($this->buildContext->getSourceDir())->trimEnd('/');
+            $sourcePath = u($file->getPathname())->after($this->buildContext->getSourceDir())->trimStart('/');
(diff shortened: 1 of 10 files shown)


OskarStark (approved):

src/CI/MissingFilesChecker.php
```diff
@@ -32,13 +33,8 @@ public function getMissingFiles(): array
...
 
         foreach ($finder as $file) {
-            $sourcePath = ltrim(substr($file->getPathname(), \strlen($this->buildContext->getSourceDir())), '/');
-
-            $htmlFile = sprintf(
-                '%s/%s.html',
-                $this->buildContext->getOutputDir(),
-                substr($sourcePath, 0, -4)
-            );
+            $sourcePath = u($file->getPathname())->after($this->buildContext->getSourceDir())->trimEnd('/');
```
nikophil, line 36: because we're doing `ltrim` shouldn't we use `trimStart`?

nikophil (commented):
i've not used that much string component so far, readability is really nice :+1:

src/CI/MissingFilesChecker.php
javiereguiluz: You are right! Fixed. Thanks!

weaverryan:
Thank you!

PR #34, opened 2021-02-26
Allow using Twig 3

javiereguiluz:


nikophil (commented):
be careful, we have a twig deprecation when php tests run:
```
1x: The spaceless tag in "paragraph.html.twig" at line 1 is deprecated since Twig 2.7, use the "spaceless" filter with the "apply" tag instead.
```

nikophil (commented):
sorry @javiereguiluz actually *our* spaceless were good: `{% apply spaceless %}`
the deprecation is coming from `doctrine/rst-parser` which uses `{% spaceless %}`

but then doesn't this mean `doctrine/rst-parser` is not compatible with twig3, and then we are not?

javiereguiluz:
@nikophil I've reverted changes.

I don't know if it's possible ... but doctrine/rst-parser should release a stable version and merge the pending PRs ... and this project should update dependencies too. The doctrine/rst-parser + docs-builder + symfony.com dependency updating is badly broken at the moment. We need stability to work on improving things. Thanks! 🙏

nikophil (approved):
ok, i see :D
https://github.com/doctrine/rst-parser/pull/130

PR #39, opened 2021-03-02
Added more properties to TOC entries

javiereguiluz:
The original TOC entry URL has something like `"url" => "dashboards.html#dashboard-route"` ... but I need each part of the URL separately because we're generating URLs with Twig's `path()` inside a Symfony app.

This PR adds the following properties to allow generating URLs with them:

```php
    "page" => "dashboards"
    "fragment" => "dashboard-route"
```

### DIFF
--- a/src/Generator/JsonGenerator.php
+++ b/src/Generator/JsonGenerator.php
@@ -89,6 +89,8 @@ class JsonGenerator
         foreach ($titles as $title) {
             $tocTree[] = [
                 'url' => sprintf('%s#%s', $metaEntry->getUrl(), Environment::slugify($title[0])),
+                'page' => u($metaEntry->getUrl())->beforeLast('.html'),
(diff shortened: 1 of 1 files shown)


src/Generator/JsonGenerator.php
```diff
@@ -89,6 +89,8 @@ private function generateToc(MetaEntry $metaEntry, ?array $titles): array
         foreach ($titles as $title) {
             $tocTree[] = [
                 'url' => sprintf('%s#%s', $metaEntry->getUrl(), Environment::slugify($title[0])),
+                'page' => u($metaEntry->getUrl())->beforeLast('.html'),
+                'fragment' => Environment::slugify($title[0]),
```
weaverryan, line 93: Bah, apparently we never created a test for this class. Ideally, we *would* have a test for this... probably most useful would be an integration test. But, the feature seems good to me 👍

PR #41, opened 2021-03-02
Changed a bit the way PHP code is highlighted

javiereguiluz:
This is an example of the endless possibilities that having a PHP-based docs builder brings to us. We can easily change or update anything.

For example, I want to highlight code in the same way that GitHub does ... which requires showing a different color for the `$` of the PHP variable names. The current highlighter doesn't support that:

image

But thanks to this super simple change, it works now:

image

### DIFF
--- a/src/Renderers/CodeNodeRenderer.php
+++ b/src/Renderers/CodeNodeRenderer.php
@@ -74,6 +74,9 @@ class CodeNodeRenderer implements NodeRenderer
         if ('text' !== $language) {
             $highLighter = new Highlighter();
             $code = $highLighter->highlight(self::LANGUAGES_MAPPING[$language] ?? $language, $code)->value;
+
+            // this allows to highlight the $ in PHP variable names
+            $code = str_replace('<span class="hljs-variable">$', '<span class="hljs-variable"><span class="hljs-variable-other-marker">$</span>', $code);
         }
 
         return $this->templateRenderer->render(
--- a/tests/fixtures/expected/blocks/code-blocks/php-annotations.html
+++ b/tests/fixtures/expected/blocks/code-blocks/php-annotations.html
@@ -27,7 +27,7 @@
 </span>
                                     {
                                     <span class="hljs-comment">/** * <span class="hljs-doctag">@Assert</span>\Iban( * message="This is not a valid International Bank Account Number (IBAN)." * ) */</span>
-                                    <span class="hljs-keyword">protected</span> <span class="hljs-variable">$bankAccountNumber</span>;
+                                    <span class="hljs-keyword">protected</span> <span class="hljs-variable"><span class="hljs-variable-other-marker">$</span>bankAccountNumber</span>;
 }
                                 </pre>
                     </div>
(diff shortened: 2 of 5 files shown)


nikophil (approved):
that's a shame the library is not configurable...

src/Renderers/CodeNodeRenderer.php
```diff
@@ -73,6 +73,9 @@ public function render(): string
         if ('text' !== $language) {
             $highLighter = new Highlighter();
             $code = $highLighter->highlight(self::LANGUAGES_MAPPING[$language] ?? $language, $code)->value;
+
+            // this allows to highlight the $ in PHP variable names
+            $code = str_replace('<span class="hljs-variable">$', '<span class="hljs-variable"><span class="hljs-variable-other-marker">$</span>', $code);
```
weaverryan, line 78: This is a hack (and I'm ok with that!). But we're going to need a test that shows that this transformation does happen. That way, when we upgrade highlightjs and they change something, we won't be surprised.

javiereguiluz:
I've updated tests ... but I don't know why this error is happening:

```
1) SymfonyDocsBuilder\Tests\IntegrationTest::testParseUnitBlock with data set "literal" ('nodes/literal')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
                                         </span>
                                     </span>
                                     {
-                                    <span class="hljs-variable">$routes</span>
+                                    <span class="hljs-variable">
+                                        <span class="hljs-variable-other-marker">$</span>
+                                        routes
+                                    </span>
                                     <span class="hljs-operator">-&gt;</span>
                                     add(
                                     <span class="hljs-string">'about_us'</span>

docs-builder/tests/IntegrationTest.php:142
```

weaverryan:
@javiereguiluz that is probably fine and you should make the "expected" test cases reflect this. Comparing HTML is SO hard due to HTML that is really the "same" but appears different due to meaningless spacing changes. To try to help with that (though it's not perfect), we use a library `gajus/dindent` to normalize the indentation and spacing on both "source" and "expected" HTML. I'm guessing that's what you're seeing: that library is putting the `routes` text onto the next line via its normalization process.

javiereguiluz:
Yes, I usually don't care about HTML spacing but in this case the generated HTML is inside a `pre` tag, so newlines are significant, right?

nikophil:
actually we're forcing the indenter to indent spans:
https://github.com/weaverryan/docs-builder/blob/master/tests/IntegrationTest.php#L306-L307

thus it is normal that we get this result.

"real" generated code looks like this:
```html
<pre class="hljs php"><span class="hljs-comment">// config/routes.php</span>
<span class="hljs-keyword">namespace</span> <span class="hljs-title">Symfony</span>\<span class="hljs-title">Component</span>\<span class="hljs-title">Routing</span>\<span class="hljs-title">Loader</span>\<span class="hljs-title">Configurator</span>;

<span class="hljs-keyword">return</span> <span class="hljs-function"><span class="hljs-keyword">function</span> <span class="hljs-params">(RoutingConfigurator <span class="hljs-variable"><span class="hljs-variable-other-marker">$</span>routes</span>)</span> </span>{
    <span class="hljs-variable"><span class="hljs-variable-other-marker">$</span>routes</span><span class="hljs-operator">-&gt;</span>add(<span class="hljs-string">'about_us'</span>, [<span class="hljs-string">'nl'</span> =&gt; <span class="hljs-string">'/over-ons'</span>, <span class="hljs-string">'en'</span> =&gt; <span class="hljs-string">'/about-us'</span>])
        <span class="hljs-operator">-&gt;</span>controller(<span class="hljs-string">'App\Controller\CompanyController::about'</span>);
};</pre>
```

so this is a false negative. We can create our expectation based on the result of the indenter, because we _know_ it's ok, it sounds a bit like fragile test.
We can also don't use the "indent span" option for this particular test...

nikophil (approved):

PR #43, opened 2021-03-03
Added a property to tweak the design of code listings

javiereguiluz:
This is very useful to tweak the design of code listings according to its length.

And I'll say it again --> even if this is a trivial change, I couldn't have made it with Sphinx in a million years. Thanks to everyone who made this PHP-based parser possible ❤️

### DIFF
--- a/src/Renderers/CodeNodeRenderer.php
+++ b/src/Renderers/CodeNodeRenderer.php
@@ -86,7 +86,7 @@ class CodeNodeRenderer implements NodeRenderer
                 // this is the number of digits of the codeblock lines-of-code
                 // e.g. LOC = 5, digits = 1; LOC = 18, digits = 2
                 // this is useful to tweak the code listings according to their length
-                'numLocDigits' => (int) floor(log10(\count($lines))) + 1,
+                'numLocDigits' => strlen((string) \count($lines)),
             ]
         );
     }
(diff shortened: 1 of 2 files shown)


src/Renderers/CodeNodeRenderer.php
```diff
@@ -83,6 +83,10 @@ public function render(): string
                 'languageMapping' => self::LANGUAGES_MAPPING[$language] ?? $language,
                 'code' => $code,
                 'lineNumbers' => rtrim($lineNumbers),
+                // this is the number of digits of the codeblock lines-of-code
+                // e.g. LOC = 5, digits = 1; LOC = 18, digits = 2
+                // this is useful to tweak the code listings according to their length
+                'numLocDigits' => (int) floor(log10(\count($lines))) + 1,
```
nikophil, line 89: i don't understand why using `log10()` here...
something like `strlen((string) \count($lines))` wouldn't be more readable?

nikophil (commented):
glad you like it :smiley:

src/Renderers/CodeNodeRenderer.php
javiereguiluz: For this kind of things I always end to go to the pure math solution ... but your proposal is more easy to understand, so I changed this. Thanks!

nikophil (approved):

weaverryan:
Very cool :)

PR #47, opened 2021-03-03
Add all the missing RST admonitions

javiereguiluz:
While reviewing the parsing of SonataAdmin docs, I saw that they use `.. important::` directive. I thought it was a mistake ... but I was the wrong one, because I didn't know that RST defines many admonitions besides the `note`, `tip`, `caution` and `warning` that we use in Symfony Docs. See https://docutils.sourceforge.io/docs/ref/rst/directives.html#specific-admonitions

So, given that adding them is trivial, I think we should add all of them. Do you agree?

### DIFF
--- a/src/Directive/AttentionDirective.php
+++ b/src/Directive/AttentionDirective.php
@@ -1,0 +1,18 @@
+<?php
+
+/*
+ * This file is part of the Docs Builder package.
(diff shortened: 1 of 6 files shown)


weaverryan:
thanks!

PR #48, opened 2021-03-03
Updated Twig code highlighter

javiereguiluz:
Updated to the latest version of the highlight.php repository.

I haven't updated the PHP on purpose because the latest version of the highlighter removes features such as highlighting PHP variables 😢


weaverryan:
> I haven't updated the PHP on purpose because the latest version of the highlighter removes features such as highlighting PHP variables 😢

🤦 

That highlights the importance of our test suite that asserts that we get the HTML we want. That'll protect us from accidentally having these things sneak up on us!

PR #51, opened 2021-03-04
Add icons to admonitions

javiereguiluz:
I'd prefer to not add these as SVG files and add them dynamically via CSS background. However, SVGs as CSS background images have lots of problems. Even `currentColor` doesn't work as stroke color (unless you do very ugly hacks).

So, let's add them as embedded files ... and if sometime we want to remove them, we can always hide them using CSS before removing them for real.


weaverryan:
Thanks Javier!

PR #56, opened 2021-03-04
WIP - fix admonition expectations

jrushlow:


javiereguiluz:
Jesse, thanks for working on this and I'm sorry I didn't take care of updating the tests myself.

jrushlow:
no worries @javiereguiluz 

The remaining failing admonition tests are due to the referenced known issue above. Should we modify the remaining tests to take this into account (without a fix)? or merge as is and create an issue / PR in the parser repo to handle whitespace. (There are probably some edges where white space is needed, e.g. innerHtml)..

I suppose we could also handle whitespace internally after the source is rendered..

tests/fixtures/expected/blocks/directives/admonition.html
```diff
@@ -4,11 +4,9 @@
     <meta charset="utf-8" />
 </head>
 <body>
-<div class="admonition-wrapper screencast_class">
-    <div class="admonition admonition-">
-        <p class="admonition-title">Screencast</p>
-        <p>Do you prefer video tutorials? Check out the the screencasts.</p>
-    </div>
+<div class="admonition admonition- screencast_class">
```
weaverryan, line 7: That looks odd. Perhaps we should make these `admonition-default` - so if you have a normal `admonition` (not a `note` or something else), we add this `-default`.

weaverryan (commented):
Let's make our tests pass (this is a minor issue anyways), and open an issue over in the rst-parser.

javiereguiluz:
@jrushlow I think the test result is correct. The extra white space comes from this Twig code that I added:

https://github.com/weaverryan/docs-builder/blob/c40871d8664a23c2876777dcf42e4ae8d9bd421f/src/Templates/default/html/directives/admonition.html.twig#L2

The "problem" is the ` {{ class ?? '' }}">` at the end of the line. Personally I think this is OK because adding this white space makes things easier for humans working on the templates, whereas machines parsing this HTML don't care about the extra white space. 

But I know this is controversial for some people, so we could change it if you want. Thanks!

tests/fixtures/expected/blocks/directives/admonition.html
javiereguiluz: Yes, it looks like a good compromise!

weaverryan:
Thanks for clarifying that Javier!

tests/fixtures/expected/blocks/directives/admonition.html
jrushlow: done

src/Templates/default/html/directives/admonition.html.twig
```diff
@@ -1,5 +1,5 @@
 {# icons are from https://heroicons.com/ - MIT License #}
-<div class="admonition admonition-{{ name }} {{ class ?? '' }}">
+<div class="admonition admonition-{{ name }} {{ class }}">
```
jrushlow, line 2: I dont think we need to check `null` here as we already set `null` be default in `AdmonitionDirective` - but i could totally be missing something obvious

https://github.com/weaverryan/docs-builder/blob/c40871d8664a23c2876777dcf42e4ae8d9bd421f/src/Directive/AdmonitionDirective.php#L27

src/Directive/AdmonitionDirective.php
```diff
@@ -20,9 +20,7 @@ public function processSub(Parser $parser, ?Node $document, string $variable, st
         $wrapperDiv = $parser->renderTemplate(
             'directives/admonition.html.twig',
             [
-                // a bit strange, but on the old markup we literally
-                // had a class of 'admonition-"
-                'name' => '',
+                'name' => 'default',
```
weaverryan, line 23: ah, cool! I forgot I did this on purpose! And now we can't do whatever we want. WEEEEEE

jrushlow: welp, that was fun.. ha :)
fixed

nikophil: Hi @jrushlow 

I don't see this update anymore?

jrushlow: I reverted the commit that added `default`.. did i misunderstand @weaverryan e.g. we must have
```
<div class="admonition-">
``` 
instead of
```
<div class="admonition-default">
```

weaverryan: Ah, sorry, I confused things. I DO want `admonition-default`.

My comment was just me realizing that I had originally done `admonition-` on purpose to exactly match the markup already used on Symfony.com. But due to how we're using this lib, we no longer need to "blindly make the same markup as we currently have on Symfony.com". We're free to make markup that is what we want and correct. So yes, please DO make the change to do `admonition-default`.

jrushlow: fixed

tests/fixtures/expected/doc-reference/index.html
```diff
@@ -4,8 +4,8 @@
     <meta charset="utf-8" />
 </head>
 <body>
-<div class="section" id="a-test-for-doc-link">
-    <h1>A test for doc link<a class="headerlink" href="#a-test-for-doc-link" title="Permalink to this headline">¶</a></h1>
+<div class="section">
+    <h1 id="a-test-for-doc-link">A test for doc link<a class="headerlink" href="#a-test-for-doc-link" title="Permalink to this headline">¶</a></h1>
```
jrushlow, line 8: @nikophil I believe this is what we want based off of https://github.com/weaverryan/docs-builder/issues/60 correct?

nikophil: this is correct, I think we can leave this PR as is and I'll fix all problems related to headings in the other PR

nikophil (approved):

Issue #46, opened 2021-03-03
Issue when parsing a head-less table


javiereguiluz:
In this file: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/docs/reference/security.rst there's this table:

```rst
The role name will be based on the name of your admin service.

========================   ======================================================
app.admin.foo              ROLE_APP_ADMIN_FOO_{PERMISSION}
my.blog.admin.foo_bar      ROLE_MY_BLOG_ADMIN_FOO_BAR_{PERMISSION}
App\\Admin\\FooAdmin       ROLE_APP\\ADMIN\\FOOADMIN_{PERMISSION}
========================   ======================================================
```

This is correctly parsed by Sphinx:

![image](https://user-images.githubusercontent.com/73419/109816437-068bd080-7c31-11eb-991c-20f72aa9e1cf.png)

But with docs-builder, the table is not parsed and I see this error message:

```
/!\ Malformed table:

app.admin.foo | ROLE_APP_ADMIN_FOO_{PERMISSION}
my.blog.admin.foo_bar | ROLE_MY_BLOG_ADMIN_FOO_BAR_{PERMISSION}
App\\Admin\\FooAdmin | ROLE_APP\\ADMIN\\FOOADMIN_{PERMISSION}
```

Tables without heads are uncommon, but I wouldn't consider this an edge-case. Thanks!

javiereguiluz:
Closing as fixed: https://github.com/weaverryan/docs-builder/pull/57#issuecomment-792678538

PR #57, opened 2021-03-05
Add support for headless tables

javiereguiluz:
This PR contains a failing test for #46.

I can't work myself on fixing this, but I thought that providing the failing test could help others fix it. Thanks!

### DIFF
--- a/tests/fixtures/expected/blocks/nodes/tables.html
+++ b/tests/fixtures/expected/blocks/nodes/tables.html
@@ -37,6 +37,22 @@
     </tr>
     <tr>
         <td>Cell 5</td>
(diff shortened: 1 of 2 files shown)


weaverryan:
This may be fixed after https://github.com/doctrine/rst-parser/pull/133 - i did massive table refactorings in that PR, which got lost. So let's check back after that's merged and we're using all of that code here.

wouterj:
I can confirm that this is fixed after updating to `"doctrine/rst-parser": "0.3.*"`.

javiereguiluz:
Fantastic! Thanks for checking it! Let's close.

Issue #45, opened 2021-03-03
Issue with <code> elements in headings


javiereguiluz:
This file: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/docs/reference/dashboard.rst has things like these:

```rst
The ``Admin`` list block
------------------------

Configuring the ``Admin`` list
------------------------------

Using the ``Admin`` service declaration
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

These are wrongly processed in the TOC, which shows the links like this:

![image](https://user-images.githubusercontent.com/73419/109800848-01715600-7c1e-11eb-984f-bbd563da4c20.png)

The contents of the heading are correctly displayed, but the "weird alphanumeric string" leaks to the permalink too:

![image](https://user-images.githubusercontent.com/73419/109801128-5745fe00-7c1e-11eb-95cb-3127033ea80b.png)

wouterj:
This is almost identical to https://github.com/doctrine/rst-parser/pull/124 (but that was fixed in another place)

nikophil:
thanks @wouterj for pointing this out :)

i've provided a fix on the parser https://github.com/doctrine/rst-parser/pull/135

javiereguiluz:
Closing as fixed. Thanks!

PR #58, opened 2021-03-05
Add support for <code> in headings

javiereguiluz:
This adds a failing test for #45 in case it's useful for the developer who tries to fix this. Thanks!


nikophil:
this needs a fix on the rst parser
https://github.com/doctrine/rst-parser/pull/135

nikophil:
I think we can close this PR, since the bug is in the rst parser, and it already has a failing test case there ;)

Issue #42, opened 2021-03-02
Unnecessary fragments in some TOC links


javiereguiluz:
In this page: https://symfony.com/doc/current/bundles/EasyAdminBundle/index.html

The TOC source code looks like this:

```html
<div class="toctree-wrapper compound">
<ul>
<li class="..."><a class="..." href="dashboards.html">Dashboards</a></li>
<li class="..."><a class="..." href="crud.html">CRUD Controllers</a></li>
<li class="..."><a class="..." href="design.html">Design</a></li>
<li class="..."><a class="..." href="fields.html">Fields</a></li>
<li class="..."><a class="..." href="filters.html">Filters</a></li>
<li class="..."><a class="..." href="actions.html">Actions</a></li>
<li class="..."><a class="..." href="security.html">Security</a></li>
<li class="..."><a class="..." href="events.html">Events</a></li>
<li class="..."><a class="..." href="upgrade.html">Upgrading from EasyAdmin 2 to EasyAdmin 3</a></li>
</ul>
</div>
```

But in the new RST parser, it looks like this:

```html
<div class="toctree-wrapper compound">
<ul>
<li><a href="dashboards.html#dashboards">Dashboards</a></li>
<li><a href="crud.html#crud-controllers">CRUD Controllers</a></li>
<li><a href="design.html#design">Design</a></li>
<li><a href="fields.html#fields">Fields</a></li>
<li><a href="filters.html#filters">Filters</a></li>
<li><a href="actions.html#actions">Actions</a></li>
<li><a href="security.html#security">Security</a></li>
<li><a href="events.html#events">Events</a></li>
<li><a href="upgrade.html#upgrading-from-easyadmin-2-to-easyadmin-3">Upgrading from EasyAdmin 2 to EasyAdmin 3</a></li>
</ul>
</div>
```

All entries have an unnecessary page fragment (e.g. `dashboards.html#dashboards` instead of just `dashboards.html`). 

The source code (see https://raw.githubusercontent.com/EasyCorp/EasyAdminBundle/master/doc/index.rst): is just:

```rst
EasyAdmin
=========

[...]

Table of Contents
-----------------

.. toctree::
    :maxdepth: 1

    dashboards
    crud
    design
    fields
    filters
    actions
    security
    events
    upgrade

[...]
```

nikophil:
Hi @javiereguiluz 

I've opened an issue on the parser about that. Let's see what they answer, otherwise I'd make it inside the docs builder

[...]

PR #59, opened 2021-03-05
Don't add unnecessary HTML fragments in TOC links

javiereguiluz:
This is a failing test for #42.

When linking to a file, we shouldn't add the HTML fragment with the `id` of the first article section. It's redundant and causes some issues in browsers when loading the page.

I can't work on fixing this right now, but this failing test will hopefully help others fix it. Thanks!

### DIFF
--- a/tests/fixtures/source/toctree/index.rst
+++ b/tests/fixtures/source/toctree/index.rst
@@ -15,6 +15,11 @@ Toctree
     directory/another_file
 
 .. toctree::
+
+    file
+    :ref:`Some specific reference <some_reference>`
+
+.. toctree::
     :glob:
 
     *
--- a/tests/fixtures/source/toctree/directory/another_file.rst
+++ b/tests/fixtures/source/toctree/directory/another_file.rst
@@ -1,2 +1,8 @@
 Another file
 ============
+
+Some text.
+
+.. _some_reference:
+
+And more text.


tests/fixtures/expected/toctree/index.html
```diff
@@ -17,12 +17,18 @@ <h1>Toctree<a class="headerlink" href="#toctree" title="Permalink to this headli
...
+</div>
+<div class="toctree-wrapper compound">
+    <ul>
+        <li><a href="directory/another_file.html">Another file</a></li>
     </ul>
 </div>
 <div class="toctree-wrapper compound">
     <ul>
-        <li><a href="directory/another_file.html#another-file">Another file</a></li>
+        <li><a href="file.html">Title</a></li>
```
nikophil, line 30: either we need to add a `:maxdepth: 1` option in the rst file, or trhe expectation file is missing 
```html
            <ul>
                <li><a href="file.html#sub-title">Sub title</a></li>
            </ul>
```

nikophil:
The fragment in the url is added by the parser. I don't know if the will accept such change, because it actually work the way it's done in doctrine's documentation.
On the other hand, I can't easily find some solution with a twig filter which removes the fragment from the first item in the toc tree... @weaverryan WDYT?

nikophil:
There is another problem: the parser does not add the link to `Some specific reference ...` in this code:
```rst
.. toctree::
    file
    :ref:`Some specific reference <some_reference>`
```

i think it's a bug in the parser, i'll fix it there

tests/fixtures/source/toctree/index.rst
```diff
@@ -14,6 +14,11 @@ Toctree
 
     directory/another_file
 
+.. toctree::
+
+    file
+    :ref:`Some specific reference <some_reference>`
```
nikophil, line 20: hi @javiereguiluz 
i can't find any examples of such nested ref inside a toctree in the whole symfony docs...

The only thing i can see which looks like this is at the bottom of this file
https://raw.githubusercontent.com/symfony/symfony-docs/5.x/components/event_dispatcher.rst
```rst
.. toctree::
    :maxdepth: 1
    :glob:

    /components/event_dispatcher/*
    /event_dispatcher/*

* :ref:`The kernel.event_listener tag <dic-tags-kernel-event-listener>`
* :ref:`The kernel.event_subscriber tag <dic-tags-kernel-event-subscriber>`
```

(there is another occurrence of the same kind of thing here: https://raw.githubusercontent.com/symfony/symfony-docs/5.x/security/auth_providers.rst)

should we then support this syntax? i'm not even sure sphinx supports it, but I may be wrong...

javiereguiluz: It's OK to remove this use case from the test then 👍  Thanks.

tests/fixtures/expected/toctree/index.html
javiereguiluz: I don't understand your comment 😐  The original RST toctree is:

```rst
.. toctree::

    directory/another_file
```

And that file doesn't have subtitles, right?

nikophil: actually this line referes to the toctree you've added:
```rst
.. toctree::

    file
    :ref:`Some specific reference <some_reference>`
```
and `file.rst` had a subtitle.
But I think we can remove this new toctree, since we decided not to handle the `:ref:` nested into a toctree

PR #66, opened 2021-03-11
Tweak the TOC Tree templates

javiereguiluz:
These new CSS classes are very useful to display differently TOCs when they are very short, very long, etc. The "deep level" also simplifies styling in some scenarios.


nikophil (approved):
LGTM

PR #71, opened 2021-03-15
Created a doc builder, doc config and doc result classes

javiereguiluz:
This parser was originally created with this target -> "integrating it in symofny.com should require the absolute minimum number of changes". That's where the PHAR and command made a lot of sense.

However, circumstances have changed completely. Now the target is "to create the best possible PHP-based RST parser and doc builder". That means less maintenance, less things to set up, less concepts to learn etc. That's why I created this PR.

The idea is that parsing RST contents would mean to run `composer require <this-package>` and then do this in your own code:

```php
$buildResult = $docBuilder->build($buildConfig);
```

Comments:

* The `BuildResult` class is a very simple value object to abstract all the build error messages.
* The `DocBuilder` is completely new ... but it's based on the command previously used to build docs.
* `BuildConfig` is a rename + simplification of the previous `BuildContext`. I think the new name is more precise, because this object mostly/only stores config values and no state/context.
* I also added some new options to `BuildConfig` which we need in `CopyImagesListener`.
 
I know this looks like a lot of changes ... but I consider it just a simplification of the current code. I thank you all for the previous work which provided a solid foundation.

Lastly, I've been using this for weeks to build bundles docs in symfony.com with success, so I consider this tested enough 🤞


src/Command/BuildDocsCommand.php
```diff
@@ -34,17 +34,17 @@ class BuildDocsCommand extends Command
...
     protected static $defaultName = 'build:docs';
 
-    private $buildContext;
+    private $buildConfig;
     private $missingFilesChecker;
     /** @var SymfonyStyle */
     private $io;
 
-    public function __construct(BuildContext $buildContext)
+    public function __construct(BuildConfig $buildConfig)
```
nikophil, line 42: What are you planning for this command? should we still keep it in order to debug or as well to be a preview tool for doc contributors?

In that case, shouldn't we leverage the new `DocBuilder` in this command?

src/DocsKernel.php
```diff
@@ -37,20 +37,20 @@ public function initBuilder(Builder $builder): void
             $builder->getErrorManager()
         );
 
-        $builder->setScannerFinder($this->buildContext->createFileFinder());
+        // $builder->setScannerFinder($this->buildConfig->createFileFinder());
```
nikophil, line 40: I think we can now un-comment this line

src/Command/BuildDocsCommand.php
javiereguiluz: Good question. I didn't want to remove the command or the PHAR-related things in this PR to minimize changes. But yes, we can remove this in the future because Symfony won't use it.

src/DocsKernel.php
javiereguiluz: Done 😅

src/Command/BuildDocsCommand.php
nikophil: don't you plan any preview tool for doc contributors?

javiereguiluz:
If possible, let's merge this soon and tag a new version because we want to start using this in symfony.com. Thanks!

src/Command/BuildDocsCommand.php
javiereguiluz: You are right. We need to do that ... but to be completely honest, that's not a priority for us right now. It's enough if this project is usable in the GitHub CI to alert us when an error happens. We can work on the "doc preview" later ... and I'd like to make it as similar as possible to the doc published on symfony.com. I think we'll be able to make that 👍

nikophil:
because the new `DocsBuilder` will be at the center of the lib, shouldn't we provide some unit/functional tests for it?

maybe before we merge @weaverryan could you review this, please?

src/Application.php
```diff
@@ -17,23 +17,12 @@
 class Application
 {
     private $application;
-    private $buildContext;
+    private $buildConfig;
 
     public function __construct(string $symfonyVersion)
```
weaverryan, line 22: We don't need this argument anymore, right? Or if we do, we should pass it directly into `BuildConfig` to set its version (which currently has a hardcoded default of 4.4)?

src/DocBuilder.php
```diff
@@ -0,0 +1,52 @@
...
+namespace SymfonyDocsBuilder;
+
+use Doctrine\RST\Builder;
+use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\Filesystem\Filesystem;
+use SymfonyDocsBuilder\CI\MissingFilesChecker;
+use SymfonyDocsBuilder\Generator\HtmlForPdfGenerator;
+use SymfonyDocsBuilder\Generator\JsonGenerator;
+
+final class DocBuilder
```
weaverryan, line 12: Unless I'm missing something, this class isn't used anywhere IN the bundle. It's just a public class that can be consumed by any project. If we DO decide to keep the `BuildDocsCommand`, I guess that command should also (in the future) use this class.

weaverryan: As far as tests go, if this is the central mechanism to build the docs, then this could/should be used in the IntegrationTest :)

weaverryan (approved):
I love the simplification! I left a few comments... that we can deal with later.

javiereguiluz:
Thanks for the review and the quick merge!

@nikophil you are right that we should test more ... but given that we are the only users of this project, I think it's reasonable to be agile/fast now (while we're integrating this in symfony.com) and then make things more robust when this integration is finished. 🙏

PR #76, opened 2021-03-17
Fixed the generation of the "prev" option in JSON generator

### LINKED ISSUE #74: The fjson file should keep the "parents" option
Sphinx describes the `parents` option as:

```
parents, prev and next:
    Information about related chapters in the TOC tree. Each relation is a
    dictionary with the keys link (HREF for the relation) and title (title
    of the related document, as HTML). parents is a list of relations,
    while prev and next are a single relation.
```

Source: https://www.sphinx-doc.org/en/1.0/builders.html#serialization-builder-details

-----

An example in practice:

This RST file from the Symfony Docs ->
https://github.com/symfony/symfony-docs/blob/5.x/contributing/code/bugs.rst

Is included in this `index.rst` from the section it belongs to:
https://github.com/symfony/symfony-docs/blob/5.x/contributing/code/index.rst

Which in turn is included in this `index.rst` from the main section:
https://github.com/symfony/symfony-docs/blob/5.x/contributing/index.rst

When the `*.fjson` is generated, it contains the following `parents`:

```json
"parents": [
    {"link": "../index.html", "title": "Contributing"},
    {"link": "index.html", "title": "Contributing Code"}
]
```

This is used in some parts of symfony.com to generate the breadcrumbs, so we'd need to re-add this option. I don't know if it's very complicated to do so or not 😐

### LINKED ISSUE #75: The `prev` option in fjson files is always null
We were checking the generated `fjson` files and we found that `next` option works perfectly, but the related `prev` option is always `null`.

Example:

This file -> https://github.com/EasyCorp/EasyAdminBundle/blob/master/doc/index.rst

The `crud` item is after `dashboards` and before `design`, but in the fjson file we get this:

```json
    "next": {
        "title": "Design",
        "link": "design.html"
    },
    "prev": null,
```

Now, if you look at `design`, you should find `crud` as `prev`. Instead, you find this:

```json
    "next": {
        "title": "Fields",
        "link": "fields.html"
    },
    "prev": null,
```

Given that `next` is working perfectly, maybe this has a non-complex fix? Thanks!

javiereguiluz:
Fixes #75.

Fixes #74 

This fixed the problem for me ... but I don't know the internals well enough, so I don't know the full implications of this change. Thanks for reviewing this proposal!

**UPDATE FROM @weaverryan**

I hijacked this PR and completely revamped the next/prev/parents functionality. It's now based exactly on how Sphinx works and it "should" be working identically.

### DIFF
--- a/src/BuildResult.php
+++ b/src/BuildResult.php
@@ -53,7 +53,7 @@ class BuildResult
     /**
      * Returns the JSON data generated for each file, keyed by the source filename.
      *
-     * @return string[]
+     * @return array[]
      */
     public function getJsonResults(): array
     {
--- a/src/BuildResult.php
+++ b/src/BuildResult.php
@@ -16,13 +16,13 @@ use SymfonyDocsBuilder\KernelFactory;
 class BuildResult
 {
     private $errors;
-    private $metas;
+    private $builder;
     private $jsonResults = [];
 
-    public function __construct(array $errors, Metas $metas)
+    public function __construct(array $errors, Builder $builder)
     {
         $this->errors = $errors;
-        $this->metas = $metas;
+        $this->builder = $builder;
     }
 
     public function appendError(string $errorMessage): void
@@ -47,7 +47,19 @@ class BuildResult
 
     public function getMetas(): Metas
     {
-        return $this->metas;
+        return $this->builder->getMetas();
+    }
+
+    /**
+     * Returns the "master document": the first file whose toctree is parsed.
+     *
+     * Unless customized, this is "index" (i.e. file index.rst).
+     *
+     * @return string
+     */
+    public function getMasterDocumentFilename(): string
+    {
+        return $this->builder->getIndexName();
     }
 
     /**
--- a/src/BuildResult.php
+++ b/src/BuildResult.php
@@ -3,6 +3,7 @@
 namespace SymfonyDocsBuilder;
 
 use Doctrine\RST\Builder;
+use Doctrine\RST\Meta\Metas;
 use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Filesystem\Filesystem;
 use SymfonyDocsBuilder\BuildConfig;
@@ -15,10 +16,13 @@ use SymfonyDocsBuilder\KernelFactory;
 class BuildResult
 {
     private $errors;
+    private $metas;
+    private $jsonResults = [];
 
-    public function __construct(array $errors)
+    public function __construct(array $errors, Metas $metas)
     {
         $this->errors = $errors;
+        $this->metas = $metas;
     }
 
     public function appendError(string $errorMessage): void
@@ -39,5 +43,25 @@ class BuildResult
     public function getErrors(): array
     {
         return $this->errors;
+    }
+
+    public function getMetas(): Metas
+    {
+        return $this->metas;
+    }
+
+    /**
+     * Returns the JSON data generated for each file, keyed by the source filename.
+     *
+     * @return string[]
+     */
+    public function getJsonResults(): array
+    {
+        return $this->jsonResults;
+    }
+
+    public function setJsonResults(array $jsonResults): void
+    {
+        $this->jsonResults = $jsonResults;
     }
 }
--- a/src/BuildResult.php
+++ b/src/BuildResult.php
@@ -15,14 +15,14 @@ use SymfonyDocsBuilder\KernelFactory;
 
 class BuildResult
 {
-    private $errors;
     private $builder;
+    private $errors;
     private $jsonResults = [];
 
-    public function __construct(array $errors, Builder $builder)
+    public function __construct(Builder $builder)
     {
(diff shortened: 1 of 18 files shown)


weaverryan:
Thanks for this Javier! I'm currently looking deeper into the toc tree handling in general - it was (I admit) once place I didn't give any serious attention to until now.

nikophil (approved):
I can't see any case where we having `'index' === $parentFile` would result into "prev" being null...

weaverryan:
I've hijacked tis PR! Description updated :)

src/DocBuilder.php
```diff
@@ -25,7 +25,10 @@ public function build(BuildConfig $config): BuildResult
         $builder = new Builder(KernelFactory::createKernel($config));
         $builder->build($config->getContentDir(), $config->getOutputDir());
 
-        $buildResult = new BuildResult($builder->getErrorManager()->getErrors());
+        $buildResult = new BuildResult(
```
javiereguiluz, line 28: We don't use the `BuildResult` class directly anywhere, so if you want to simply pass the `$builder` as argument, it's OK to do so:

```php
$buildResult = new BuildResult($builder);
```

src/Generator/JsonGenerator.php
```diff
@@ -37,28 +37,49 @@ public function __construct(Metas $metas, BuildConfig $buildConfig)
...
             $data = [
                 'title' => $metaEntry->getTitle(),
+                'parents' => $this->determineParents($parserFilename, $tocTreeHierarchy) ?: [],
                 'current_page_name' => $parserFilename,
                 'toc' => $this->generateToc($metaEntry, current($metaEntry->getTitles())[1]),
-                'next' => $this->guessNext($parserFilename),
-                'prev' => $this->guessPrev($parserFilename),
+                'next' => $next,
+                'prev' => $prev,
                 'rellinks' => [
```
javiereguiluz, line 79: This is the official Sphinx explanation of `rellinks`: https://www.sphinx-doc.org/en/master/templating.html#rellinks

I'd leave this as it is because I think we don't use this option anywhere.

tests/fixtures/source/json/crud.rst
```diff
@@ -0,0 +1,7 @@
+CRUD
+====
+
+CRUD stands for: "create, read, update, delete".
+
+Or it might be: "Crazy runner's utter delight" (which would
```
javiereguiluz, line 6: 😂

tests/fixtures/source/json/fields.rst
```diff
@@ -0,0 +1,5 @@
+Fields
+======
+
+I love fields: big open prairies, grass fields, corn fields... reall,
```
javiereguiluz, line 4: `reall` -> `really` ?

javiereguiluz (commented):
Ryan, thanks for this great contribution!

I love how you fixed everything, while simplifying and improving everything at the same time. Stellar work! 🙇‍♂️

src/BuildResult.php
```diff
@@ -40,4 +44,36 @@ public function getErrors(): array
     {
         return $this->errors;
     }
+
+    public function getMetas(): Metas
```
fabpot, line 48: Not directly related to this PR, but I suppose that `getMetadata()` would be more correct (which should also be modified in the builder).

src/BuildResult.php
```diff
@@ -40,4 +44,36 @@ public function getErrors(): array
...
+    {
+        return $this->builder->getMetas();
+    }
+
+    /**
+     * Returns the "master document": the first file whose toctree is parsed.
+     *
+     * Unless customized, this is "index" (i.e. file index.rst).
+     *
+     * @return string
```
fabpot, line 58: Could removed as the PHP signature has the same. Not sure what's the policy in this repo though.

src/BuildResult.php
```diff
@@ -40,4 +44,36 @@ public function getErrors(): array
...
+     */
+    public function getMasterDocumentFilename(): string
+    {
+        return $this->builder->getIndexName();
+    }
+
+    /**
+     * Returns the JSON data generated for each file, keyed by the source filename.
+     *
+     * @return array[]
```
fabpot, line 68: If description is good, this type hint looks wrong.

src/Generator/JsonGenerator.php
fabpot: As we already have changed the format of the file anyway, we could remove rellinks support as we don't use it anywhere AFAIK.

fabpot (commented):
Some random comments

src/Generator/JsonGenerator.php
weaverryan: I have removed it. If there is other information that would be useful, let me know and we an add it :)

src/BuildResult.php
weaverryan: Yes, thank you - my editor added that, it was not my intention.

PR #77, opened 2021-03-18
Updated literal blocks to improve word breaking

javiereguiluz:
In Symfony Docs is very common to have long PHP namespaces as `<code>` elements.

Firefox is the smartest browser and can break them by the backslash:

firefox-default

However, Chrome and others don't do that, so the result is large and ugly gaps in the text:

chrome-default

This PR proposes a *trick* to add `<wbr>` tags in some cases to help browsers. These `<wbr>` are completely transparent to users and they don't appear even if you copy+paste the text.

After this PR, Chrome looks like this:

chrome-with-wbr

### DIFF
--- a/src/Renderers/SpanNodeRenderer.php
+++ b/src/Renderers/SpanNodeRenderer.php
@@ -62,7 +62,21 @@ class SpanNodeRenderer extends BaseSpanNodeRenderer
         );
     }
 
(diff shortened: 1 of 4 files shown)


wouterj (approved):

weaverryan:
Absolutely gorgeous!

PR #78, opened 2021-03-18
Adding a span around code blocks that start with $

### LINKED ISSUE #68: Mark prompts differently in terminal/console blocks
`code-block:: terminal` is highlighted like this:

image

We need the "prompt" (`$` in Linux and macOS and `C:\>` in Windows) to be marked differently. This way we can apply different CSS styles to it and make it non-selectable by users (important when copying commands).

Today we do this using some ugly JS/CSS hacks ... but thanks to the new RST parser we could do this in PHP while building docs. It's similar to this -> https://github.com/weaverryan/docs-builder/pull/41/files

We could use `hljs-prompt` as the CSS class of this new element. Any volunteer to take care of this? Thanks!

weaverryan:
Fixes #68

### DIFF
--- a/src/Renderers/CodeNodeRenderer.php
+++ b/src/Renderers/CodeNodeRenderer.php
@@ -74,6 +74,10 @@ class CodeNodeRenderer implements NodeRenderer
             $highlightedCode = str_replace('<span class="hljs-variable">$', '<span class="hljs-variable"><span class="hljs-variable-other-marker">$</span>', $highlightedCode);
         }
 
+        if ('terminal' === $language && str_starts_with($highlightedCode, '$')) {
+            $highlightedCode = '<span class="hljs-prompt">$</span>' . substr($highlightedCode, 1);
+        }
+
         $numOfLines = \count(preg_split('/\r\n|\r|\n/', $highlightedCode));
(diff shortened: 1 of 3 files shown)


wouterj:
Let's make it a bit more difficult Ryan, the python terminal lexer we're using also takes care of some other (windows & comments) bash highlights: https://github.com/symfony/symfony-docs/blob/5.x/_build/_exts/symfonycom/sphinx/lexer.py

weaverryan:
> Let's make it a bit more difficult Ryan

Oh, boooooo. You got me Wouter! 😆

weaverryan:
Ok, should be better now. This may not imitate exactly what the old one does. But... we can basically do whatever we want. The test case holds all the scenarios and markup... and we can add more - https://github.com/weaverryan/docs-builder/pull/78/files#diff-ffe40345fc14908fa9ef0846da05ce80b3398903e81e02c5363c21451346adea

javiereguiluz (approved):
Nice one! No more CSS/JS hacks 🙈

weaverryan:
Yay! And if we need further tweaks, as you can see, it should be easy :)

PR #80, opened 2021-03-19
[WIP] Fix missing line break with anchors

### LINKED ISSUE #37: Minor issue when parsing a reference
In this article -> https://raw.githubusercontent.com/EasyCorp/EasyAdminBundle/master/doc/fields.rst we have this:

```rst
Next, create the template used to render the field in the ``index`` and ``detail``
:ref:`CRUD pages <crud-pages>`. The template can use any `Twig templating features`_
and the following variables:
```

Everything is correctly parsed, but there's a minor issue. The generated HTML is:

```html
<p>Next, create the template used to render the field in the <code>index</code> and <code>detail</code><a href="crud.html#crud-pages" class="reference internal">CRUD pages</a>. The template can use any <a href="https://twig.symfony.com/doc/3.x/" class="reference external">Twig templating features</a>
```

The problem is that there isn't a white space between `<code>detail</code>` and the following `<a>` reference. The missing white space is needed to avoid this:

image

weaverryan:
This fixes #37 

But... the fix should maybe go upstream to docrtrine... and I don't quite understand it. I included some test normalization in this PR, but the ACTUAL bug fix was to copy the `paragraph.html.twig` template from `rst-parser` into our project and customize it. Here are the changes:

```diff
-{% apply spaceless %}
    {% set text = paragraphNode.value.render()|trim %}

    {% if text %}
        <p{% if paragraphNode.classes %} class="{{ paragraphNode.classesString }}"{% endif %}>{{ text|raw }}</p>
    {% endif %}
-{% endapply %}
```

Yes, the `spaceless` was having bad side effects. What would happen is that the `text` variable would be set to something like this (with line breaks):

```
<p>Next, create the template with <code translate="no" class="notranslate">details</code>
<a href="https://symfony.com/doc/current/bundles/EasyAdminBundle/crud.html" class="reference external">CRUD pages</a>.</p>
```

The line break caused the word `details` and `CRUD` to have a space between them. But the `spaceless` killed that.

Are we somehow misusing / abusing spaceless? Should we remove all the spaceless upstream in the Doctrine templates?

Silly spacing thing in test depends on https://github.com/doctrine/rst-parser/pull/141

Cheers!


javiereguiluz:
Thanks for taking care of this one.

About spaces ... I don't know if it's possible, but my opinion would be to never deal with spaces manually unless it's required (e.g. inside a `<code>` or `<pre>` because spaces are significant or inside a `<a>` because additional spaces are visible if you use text underline).

weaverryan:
> About spaces ... I don't know if it's possible

@javiereguiluz do you mean that you would prefer that the templates never use the `spaceless` filter? Or are you saying something else?

Thanks!

javiereguiluz:
Yes, I meant that ... but maybe now that we already use `spaceless`, it's smarter to just remove it where it makes sense and keep the rest.

weaverryan:
Ok, for now, let's handle this in our repo, removing spaceless on a case-by-case basis :)

stof:
Spaceless is about removing spaces between sibling HTML tags. but when these tags are inline elements, the whitespaces are relevant to the rendering.
Removing the spaceless filtering on a case by case basis would also require removing in any parent in the rendering.

In my experience, lots of cases using spaceless are using it to remove the template indentation, which is a misuse for 2 reasons:
- the indentation might not be between 2 tags (see for instance the code snippet in this issue, where the indentation is before the first tag of the filtered content)
- it removes much more spaces (it can remove spaces in the content rendered by the `text` variable, which is the case here)

The proper solution to remove spaces introduced by template indentation is to use the Twig whitespace control feature instead (which also has the nice performance benefit of being a compile-time feature rather than a runtime one)

PR #82, opened 2021-03-19
Consider "versionadded" as an admonition and improve "seealso" admonition

javiereguiluz:
This PR makes two changes:

* Handles/renders "versionadded" directives as adominitions, to minimize the changes needed when publishing contents.
* It adds an icon to "seealso" admonition to better match the style of other admonitions.

In practice it looks like this:

image

### DIFF
--- a/src/Directive/VersionAddedDirective.php
+++ b/src/Directive/VersionAddedDirective.php
@@ -23,12 +23,15 @@ class VersionAddedDirective extends SubDirective
     public function processSub(Parser $parser, ?Node $document, string $variable, string $data, array $options): ?Node
     {
         $wrapperDiv = $parser->renderTemplate(
-            'directives/version-added.html.twig',
+            'directives/admonition.html.twig',
             [
+                'name' => 'versionadded',
+                'text' => sprintf('New in version %s', $data),
+                'class' => $options['class'] ?? null,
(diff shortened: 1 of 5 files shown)


src/Directive/VersionAddedDirective.php
```diff
@@ -23,12 +23,15 @@ public function getName(): string
     public function processSub(Parser $parser, ?Node $document, string $variable, string $data, array $options): ?Node
     {
         $wrapperDiv = $parser->renderTemplate(
-            'directives/version-added.html.twig',
+            'directives/admonition.html.twig',
```
weaverryan, line 26: The `version-added.html.twig` file can be deleted now, right?

weaverryan: Or is the old template still used somehow?

javiereguiluz: Yes, sorry! I forgot about that 😞   Please, delete it if you can. Thanks!

weaverryan: I did that, and the tests fail! Something else is going on...

weaverryan: nevermind - it's fine - it was my mistake, I'm deleting it now :)

PR #84, opened 2021-03-20
Fix links in generated JSON

fabpot:
* Links are relative to the current file
* `parents` is an array of links like the ones for `prev` and `next`


weaverryan:
Wonderful! Thank you!

PR #85, opened 2021-03-23
Tweaked the configuration block directive

javiereguiluz:
Configuration blocks weren't styled yet in the new doc builder, so I did that and to make CSS and JS code much easier, I propose this tweaks in the generated HTML code for them.

### DIFF
--- a/tests/fixtures/expected/main/datetime.html
+++ b/tests/fixtures/expected/main/datetime.html
@@ -191,9 +191,9 @@ with the <a href="datetime.html#date-widget" class="reference internal">date_wid
 <p>And configuration blocks:</p>
 <div class="configuration-block">
     <ul class="configuration-tabs">
-        <li> <a href="#" data-active data-language="YAML">YAML</a> </li>
-        <li> <a href="#" data-language="XML">XML</a> </li>
-        <li> <a href="#" data-language="PHP">PHP</a> </li>
+        <li data-language="YAML" data-active="true"> <a href="#">YAML</a> </li>
+        <li data-language="XML" > <a href="#">XML</a> </li>
+        <li data-language="PHP" > <a href="#">PHP</a> </li>
     </ul>
     <div class="configuration-codeblock" data-language="YAML" style="">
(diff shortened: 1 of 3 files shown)


javiereguiluz:
Sorry for the continuous changes in the proposed HTML code. I'm happy with the latest result, so I don't plan to change this for now ... so this can be reviewed. Thanks!

tests/fixtures/expected/main/datetime.html
```diff
@@ -190,10 +190,13 @@ <h3 id="error-bubbling"><a class="headerlink" href="#error-bubbling" title="Perm
...
 <p>And configuration blocks:</p>
 <div class="configuration-block">
-    <ul class="simple">
-                    <li>
-                <em>YAML</em>
-                <div translate="no" class="notranslate codeblock codeblock-loc-1 codeblock-yaml">
+    <ul class="configuration-tabs">
+        <li data-language="YAML" data-active="true"> <span>YAML</span> </li>
+        <li data-language="XML" > <span>XML</span> </li>
+        <li data-language="PHP" > <span>PHP</span> </li>
```
OskarStark, line 196:
```diff
        <li data-language="PHP" ><span>PHP</span></li>
```
Do we really need to keep these spaces?

javiereguiluz: I'd prefer to keep it (it's the same case as https://github.com/weaverryan/docs-builder/pull/56#issuecomment-791463260).

I know some people disagree but ... I prefer to make templates readable instead of making the generated HTML perfect, because templates are for humans (and they must maintain them in the future) and HTML is for machines and they don't care.

I hope you agree somewhat on this 🙏

nikophil (approved):

tests/fixtures/expected/main/datetime.html
nikophil: :+1: to be pragmatic on this kind of things

weaverryan:
Thanks Javier!

PR #86, opened 2021-03-23
Fixed the handling of custom image prefix

javiereguiluz:
I introduced some issues when refactoring this. The new proposed logic is:

* If you provide a custom "image prefix", then the generated image URL is `<your prefix> + / + <image relative path>` (we don't change that URL to a relative one to avoid issues ... we use your provided prefix "as is")
* If you don't provide this custom "image prefix", we consider that images are stored in `<output dir>/_images/` and calculate image URLs as relative URLs.

### DIFF
--- a/src/Listener/CopyImagesListener.php
+++ b/src/Listener/CopyImagesListener.php
@@ -53,7 +53,11 @@ class CopyImagesListener
         $newAbsoluteFilePath = $this->buildConfig->getImagesDir().'/'.$fileInfo->getFilename();
         $fs->copy($sourceImage, $newAbsoluteFilePath, true);
 
-        $newUrlPath = $this->buildConfig->getImagesPublicPrefix().'/'.$fileInfo->getFilename();
-        $node->setValue($node->getEnvironment()->relativeUrl($newUrlPath));
+        if ('' === $this->buildConfig->getImagesPublicPrefix()) {
+            $newUrlPath = $node->getEnvironment()->relativeUrl('_images/'.$fileInfo->getFilename());
+        } else {
+            $newUrlPath = $this->buildConfig->getImagesPublicPrefix().'/'.$fileInfo->getFilename();
+        }
+        $node->setValue($newUrlPath);
     }
 }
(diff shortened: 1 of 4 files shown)


src/Listener/CopyImagesListener.php
```diff
@@ -53,7 +53,11 @@ public function preNodeRender(PreNodeRenderEvent $event)
         $newAbsoluteFilePath = $this->buildConfig->getImagesDir().'/'.$fileInfo->getFilename();
         $fs->copy($sourceImage, $newAbsoluteFilePath, true);
 
-        $newUrlPath = $this->buildConfig->getImagesPublicPrefix().'/'.$fileInfo->getFilename();
-        $node->setValue($node->getEnvironment()->relativeUrl($newUrlPath));
+        if ('' === $this->buildConfig->getImagesPublicPrefix()) {
+            $newUrlPath = $node->getEnvironment()->relativeUrl('_images/'.$fileInfo->getFilename());
+        } else {
+            $newUrlPath = $this->buildConfig->getImagesPublicPrefix().'/'.$fileInfo->getFilename();
```
nikophil, line 59: Hii @javiereguiluz 

we're copying the image to `$this->buildConfig->getImagesDir().'/'.$fileInfo->getFilename()` but we're linking to `$this->buildConfig->getImagesPublicPrefix().'/'.$fileInfo->getFilename()`

isn't there a problem if these two paths are uncorrelated?

in the test below, the img's src is `/some/custom/prefix-for-images/symfony-logo.png` but the image file path is `./_images/symfony-logo.png`

javiereguiluz: You are right. However, this is intended.

The default behavior is unchanged (copy images in `<contents/_images/` and link to them in that dir using a relative link). However, for advanced/custom setups, the images are copied to a separate dir (or even server/service like AWS). The missing piece that you mentioned is that in your own code using this Doc Builder you must take care of this and copy/move the images accordingly.

weaverryan:
A test is included... so I'm happy with this if it gives us the end result we need :)

PR #88, opened 2021-03-24
Made the counting of code lines more robust

javiereguiluz:
I saw some issues in the counting of lines in some generated docs ... so this PR tries to make things more robust.

As you can see in tests, we remove a trailing new line in some code blocks. This is correct, because new lines inside `<pre>` are significant. So, this is a bug fix in those cases.

**Update**: the following error was fixed thanks to Christophe's suggestion. Thanks!

~However, there's a problem with these tests:~

* ~`tests/fixtures/expected/blocks/code-blocks/html-twig.html`~
* ~`tests/fixtures/expected/blocks/code-blocks/twig.html`~

~I had to update those tests to fix them, but this is wrong. The extra line added to the generated HTML is wrong. I can't see why this happens. Can anyone help me? Thanks!~

### DIFF
--- a/src/Renderers/CodeNodeRenderer.php
+++ b/src/Renderers/CodeNodeRenderer.php
@@ -79,8 +79,9 @@ class CodeNodeRenderer implements NodeRenderer
             $highlightedCode = preg_replace('/^C:\\\&gt;/m', '<span class="hljs-prompt">C:\&gt;</span>', $highlightedCode);
         }
 
+        $highlightedCode = trim($highlightedCode);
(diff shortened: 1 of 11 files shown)


stof:
I think you need to trim the codeblock before highlighting rather than after. In this html-twig.html, the final LF char is before the `</span>` of the hilighting, not after the span, and so won't be trimmed. This indeed keeps 3 lines in the source for the highlighted code

javiereguiluz:
@stof your suggestion fixed the problem! Thanks a lot!

nikophil (approved):

weaverryan:
Awesome - thanks everyone!

PR #91, opened 2021-03-31
Added some missing labels in configuration blocks

javiereguiluz:
In addition to the main fix, I made another change in this PR.

The current labels for "php-symfony" and "php-standalone" are "Framework Use" and "Standalone Use". I don't think they are bad, but they could be better.

In this PR I propose to use "Symfony apps" and "Standalone PHP apps". Here's a comparison:

image

I like "Symfony apps" a lot but I don't like "Standalone PHP apps" the same. Maybe "Other PHP apps", "Non-Symfony apps", ... ? Or maybe the original labels are better and we should keep them?

What do you think @weaverryan, @wouterj? Thanks!


src/Directive/ConfigurationBlockDirective.php
```diff
@@ -17,6 +17,27 @@
 
 class ConfigurationBlockDirective extends SubDirective
 {
+    private const LANGUAGE_LABELS = [
+        'env' => 'Bash',
```
OskarStark, line 21: I would rather do sth. like:

```suggestion
    private const LANGUAGES = [
        [
            'type' => 'env',
            'label' => 'Bash',
        ],
```

In this case it will be much easier to extend it later if we need some special settings

javiereguiluz: If this was a public package, I'd agree ... but since this is internal and only for Symfony Docs, we can change anything at any time (BC breaks don't apply to us in this project). So, I'd keep it like this.

Also, in other classes we have similar, but different, config -> https://github.com/weaverryan/docs-builder/blob/4080f308aa6359da1e9643ea18faf9d2c52daf9a/src/Renderers/CodeNodeRenderer.php#L24 so we should refactor everything to have a common config ... but Id' keep it like it is for now.

javiereguiluz:
Please, don't forget to add your suggestions here. Thanks!

javiereguiluz:
I've reverted the proposed label renames to make this mergeable and we can discuss about renaming in the future.

PR #93, opened 2021-04-06
Add a README file

javiereguiluz:
Please review the proposed contents of this file. I don't want to sound harsh, but I'd like to make it extremely clear that this is an internal tool that should be used only by Symfony.


OskarStark (approved):

jrushlow (approved):
Doesn't sound harsh at all. Clear and to the point. It may be worth adding that support is not provided if someone uses this etc..

nikophil (approved):

PR #95, opened 2021-04-08
refacto: don't extend from parser

nikophil:
Hello,

this is a solution in response to this PR
https://github.com/doctrine/rst-parser/pull/144

I used composition over inheritance for both classes where we was extending parser's classes.
This way @greg0ire could cleanly close the parser's api without exceptions.

### DIFF
--- a/src/Renderers/SpanNodeRenderer.php
+++ b/src/Renderers/SpanNodeRenderer.php
@@ -14,35 +14,34 @@ namespace SymfonyDocsBuilder\Renderers;
 use Doctrine\RST\Environment;
 use Doctrine\RST\HTML\Renderers\SpanNodeRenderer as BaseSpanNodeRenderer;
 use Doctrine\RST\Nodes\SpanNode;
-use Doctrine\RST\Templates\TemplateRenderer;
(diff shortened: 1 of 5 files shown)


src/Renderers/SpanNodeRenderer.php
```diff
@@ -53,18 +52,25 @@ public function link(?string $url, string $title, array $attributes = []): strin
...
+            $reference = new ResolvedReference(
+                $reference->getFile(),
+                $reference->getTitle(),
+                $reference->getUrl(),
+                $reference->getTitles(),
+                $this->addAttributesForUnsafeUrl($reference->getAttributes())
+            );
+        }
+
+        return $this->decoratedSpanNodeRenderer->reference($reference, $value);
```
nikophil, line 73: this part was a bit tricky because `BaseSpanNodeRenderer::reference()` internally calls `BaseSpanNodeRenderer::link()` thus I had to update the attributes here inside the reference

weaverryan:
Wow, yea, a bit complex on the `SpanNodeRenderer` - that is complex enough already! But, tests pass and the change looks reasonable. Thanks!

PR #96, opened 2021-04-16
Use a released version of doctrine/rst-parser

Nyholm:


wouterj (approved):

javiereguiluz (approved):

composer.json
```diff
@@ -16,7 +16,7 @@
     "require": {
         "ext-json": "*",
         "ext-curl": "*",
-        "doctrine/rst-parser": "0.4.x-dev",
+        "doctrine/rst-parser": "^0.4",
```
jrushlow, line 19: would it be better todo `>=0.4` so we dont have to do a version bump when they push out `0.5`? i'm not sure of the over stability of rst-parser (even in dev)..

Nyholm: Good question!
If they push 0.5 we would like to review all the changes carefully before we use the new version. 

Semver works differently before 1.0. Basically, minor versions can include BC breaks.

jrushlow (approved):

weaverryan:
Thanks Tobias!

PR #97, opened 2021-04-16
Validate code blocks

Nyholm:
Make sure that all our code examples are using valid syntax. I found quite a few things that should be updated. 

This PR does not do any style checks, it just makes sure things can be parsed. 

Note that validating the PHP examples are significantly slower. To build symfony docs takes 164s compared to 23s before this patch. Im sure there are optimisations and caching one can do. But I just wanted to share the performance impact.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "symfony/css-selector": "^5.2",
         "symfony/console": "^5.2",
         "twig/twig": "^2.14 || ^3.3",
-        "symfony/http-client": "^5.2"
+        "symfony/http-client": "^5.2",
+        "symfony/yaml": "^5.2"
     },
     "require-dev": {
         "gajus/dindent": "^2.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "2739ab0cffd3ad3caba1283d0b2e09e2",
+    "content-hash": "c36d501ec8535171300bfa3773f593bd",
     "packages": [
         {
             "name": "doctrine/event-manager",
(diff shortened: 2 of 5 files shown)


OskarStark:
Now as I can see the code it could be placed in DOCtor as a simple rule.

Not sure you want to give it a try

Nyholm:
I saw that all syntax checks was missing from this project. 

Once this is merged I will start to verify code blocks in a smarter way. Ie, when we define PHP Config with the ConfigBuilders I want to make sure the methods exists etc. So I need to side-load a Symfony project (with the correct version) to make sure things are working. 

I can do the same with Yaml and XML config too. But this PR is a first step. 

Im not sure if this is a feature of `DOCtor` or this library. Im happy to discuss that with the Symfony docs team.

wouterj:
Oh, I definitely think this does belong to DOCtor more than this library (let's keep this library focused at building all the docs in symfony.com's infrastructure, that's already a pretty big task for 1 library).

However, I think the Doctrine rst parser library opens up a whole new world for DOCtor. As Doctrine takes care of all the parsing, DOCtor can be more powerful (by creating event listeners on specific nodes, instead of first trying to make sense of plain text rst and then reviewing it). This PR proves how "easy" it is to create new check rules. This completely bypasses my previous attempts on this topic (https://github.com/OskarStark/doctor-rst/pull/719).

wouterj:
> Once this is merged I will start to verify code blocks in a smarter way. Ie, when we define PHP Config with the ConfigBuilders I want to make sure the methods exists etc. So I need to side-load a Symfony project (with the correct version) to make sure things are working.

While discussing to try out the Doctrine rst parser back in 2019, we also talked about this (smarter code checking, and even generation). Maybe we can learn from Fabien's way of building the book and also have some sort of "auto playable" code examples. In any case, I think it's a bit of a big topic to just start now to be honest. I would love to see all code examples to be updated to ConfigBuilders when 5.3.0 is released, and having this complex lexer first will make this near impossible I think.

Nyholm:
I just started looking into this library and DOCtor. As currently see it, it PR would be really hard to do without `doctrine/rst-parser`. I also agree with you that the PR does not belong in this library. Here we should focus on building stuff, not validating. 

@OskarStark, do you see that we can fit in this into DOCtor? I was looking at some of the existing rules and they all work with plain texts, line by line. 

The alternative is to create a new "Code block verifier" project. That means a new CI job in symfony/symfony-docs. 


> I would love to see all code examples to be updated to ConfigBuilders when 5.3.0 is released, and having this complex lexer first will make this near impossible I think.

Im unavailable tomorrow, but I'll can fix the parsers and a "baseline feature" on Sunday. ;)

Nyholm:
Im closing this PR. I added a small tool here: https://github.com/symfony-docs-tools/code-block-checker

See this PR: https://github.com/symfony/symfony-docs/pull/15257

PR #98, opened 2021-04-19
Add platform check file to phar

gseidel:
If I compile the phar file with `bin/compile` and execute the `docs.phar` file later. I got the error 

```
 failed to open stream: phar error: "vendor/composer/platform_check.php" is not a file in phar
 ```

Tested with latest composer version `2.0.12`


weaverryan:
Thanks @gseidel!

FYI - we may remove the phar functionality. But until then, I'm happy to merge any fixes.

PR #99, opened 2021-04-23
Update composer package organization

Nyholm:
This will sync up with https://github.com/symfony-tools/code-block-checker/pull/35.

It will allow two things: 

1) We can properly use a dependency in https://github.com/symfony-tools/code-block-checker/pull/39
2) We could (if we wanted to) remove the composer.json in https://github.com/symfony/symfony-docs/pull/15284

When this PR is merged, I'll add the package on packagist.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "symfony/docs-builder",
+    "name": "symfony-tools/docs-builder",
     "description": "The build system for Symfony's documentation",
     "type": "project",
     "license": "MIT",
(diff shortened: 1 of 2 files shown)


wouterj:
Hmm, I'm not so sure about this. Is there any advantages to publishing this on packagist, rather than using the current repositories option?

Not publishing on Packagist communicates very clearly that this is an open source internal tool that is not meant to be used outside Symfony's CI/CD.

weaverryan (approved):
Your argument makes sense to me too. However, I think Fabien is in favor of this change... and... if we have a clear warning on this library's readme (which we already do), and people still use it, then they know they are "on their own". So it's a 👍 from me

javiereguiluz (approved):
I agree with Wouter ... but I'm voting +1 to try to move this forward.

We can always remove this project from Packagist and/or GitHub in the future if we want. Thanks!

OskarStark (approved):

PR #100, opened 2021-04-23
Opt-in to printing errors to standard output

Nyholm:
This is technically a BC break, but since it is an internal tool ¯\_(ツ)_/¯

Symfony-docs shows the errors: 

https://github.com/symfony/symfony-docs/blob/5.x/.github/workflows/ci.yaml#L76-L78

### DIFF
--- a/src/Command/BuildDocsCommand.php
+++ b/src/Command/BuildDocsCommand.php
@@ -91,6 +91,12 @@ class BuildDocsCommand extends Command
                 InputOption::VALUE_NONE,
                 'Return a non-zero code if there are errors/warnings'
             )
+            ->addOption(
+                'print-errors',
+                null,
+                InputOption::VALUE_NONE,
+                'Print errors when to standard output'
+            )
         ;
     }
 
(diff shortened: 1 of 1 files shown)


src/Command/BuildDocsCommand.php
```diff
@@ -91,6 +91,12 @@ protected function configure()
                 InputOption::VALUE_NONE,
                 'Return a non-zero code if there are errors/warnings'
             )
+            ->addOption(
+                'print-errors',
+                null,
+                InputOption::VALUE_NONE,
+                'Print errors when to standard output'
```
weaverryan, line 98:
```suggestion
                'Print errors to standard output'
```

Maybe move this up under `save-errors`?

Or should we simply use the "verbosity" check - like maybe we print errors with `-v`? Or should we continue printing them... but NOT print them if you pass a `-q` for quiet? Sphinx DOES show errors while it's building (or, at least at the end... I can't remember exactly when it shows them)

wouterj: > Or should we continue printing them... but NOT print them if you pass a -q for quiet?

:+1: for this behavior. 

This is indeed consistent with Sphinx (which indeed shows errors directly while building).

Nyholm: Hm.. But we build the `-vvv` argument. So I dont think it should be related to verbosity. 

> This is indeed consistent with Sphinx

That is true, but is that important?

wouterj:
> Hm.. But we build the -vvv argument. So I dont think it should be related to verbosity.

I'm almost 100% sure that I added `-vvv` as a desperate move to try to get error output (I added the CI pipeline with 0 config of this tool). I wouldn't give any weight to the docs CI pipeline in any decision about this tool.

> That is true, but is that important?

Not really, but it's a good reference (it's the most commonly used documentation build tool, so if it does something, someone probably thought about it and wanted that feature). I think it makes sense to always show errors.

Nyholm:
I've updated the PR. It will now just add an option to disable parsing errors on std-out.

PR #102, opened 2021-07-02
Fixed the generation of TOCs

javiereguiluz:
In some private tests we saw that a short TOC was displayed in two columns (as if it was a long TOC) whereas a slightly longer TOC was displayed in a single column.

The reason is that we were counting all TOC elements, instead of only the visible elements.

Please note that this PR is almost, but not entirely, correct. I had to hardcode the number of visible TOC levels at `2` (by far, the most common value in Symfony-related docs) but this should instead be obtained dynamically via the `maxdepth` TOC tree option in the RST contents.

### DIFF
--- a/src/Twig/TocExtension.php
+++ b/src/Twig/TocExtension.php
@@ -24,10 +24,16 @@ class TocExtension extends AbstractExtension
     public static function getOptions(array $toc): array
     {
         $flattendToc = self::flattenToc($toc);
-        $maxDepth = 0;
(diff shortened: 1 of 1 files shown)


Nyholm (approved):
Im happy with this. I opened an issue to resolve this TODO. 

https://github.com/symfony-tools/docs-builder/issues/106

weaverryan:
Thanks Javier!

PR #108, opened 2021-07-07
Fix a formatting issue in definition lists

javiereguiluz:
This PR adds more tests related to lists and **adds a failing test** for some bug that I've found. I don't know how to solve this, so **I need your help**. Thanks!

-----

In the docs of LiipImagineBundle I've found this error. Above you see the original RST content and below the rendered HTML, which is wrong:

The `** ... **` should instead be a bold text `<strong> ... </strong>`.

The failing test shows this problem well:

```diff
--- Expected
+++ Actual
@@ @@
 <dl>
-    <dt><strong>Quis:</strong> <code translate="no" class="notranslate">int[]</code></dt>
+    <dt>
+        **Quis
+        <span class="classifier-delimiter">:</span>
+        <span class="classifier">** <code translate="no" class="notranslate">int[]</code></span>
+    </dt>
```

### DIFF
--- a/tests/IntegrationTest.php
+++ b/tests/IntegrationTest.php
@@ -124,8 +124,20 @@ class IntegrationTest extends AbstractIntegrationTest
             'blockName' => 'nodes/literal',
         ];
 
-        yield 'list' => [
-            'blockName' => 'nodes/list',
+        yield 'unordered-list' => [
+            'blockName' => 'nodes/unordered-list',
+        ];
+
+        yield 'ordered-list' => [
+            'blockName' => 'nodes/ordered-list',
+        ];
+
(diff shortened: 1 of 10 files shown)


wouterj:
This is now fixed in version 0.4.1 of the rst-parser library. I wondering if it makes sense to duplicate syntax related tests in this package? (i.e. merge this PR)

javiereguiluz:
First, thanks a lot for fixing this bug in the upstream repository 🙇 

Yes, let's close this PR because it only adds tests for things that should already be tested in the Doctrine RST parser library. For the future, we might want to rethink the tests of this project to remove everything that is already tested upstream. Thanks!

PR #109, opened 2021-07-09
Allow to build RST content strings

javiereguiluz:
The common use case is to build all the RST contents stored in some dir. However, sometimes you need to build just the contents of a single file or a string. For example we need this for symfony.com, where many blog posts are written as RST contents.

This PR adds a new `buildString(string $contents): BuildResult` content to complement the current `build(BuildConfig $config): BuildResult` method.

I had to add some config options and methods ... please tell me what you think about their naming. Thanks!


src/BuildConfig.php
```diff
@@ -30,6 +30,8 @@ class BuildConfig
     private $subdirectoryToBuild;
     private $excludedPaths;
     private $fileFinder;
+    private $isContentAString;
```
OskarStark, line 33:
```suggestion
    private $contentIsString;
```
?

PR #111, opened 2021-08-22
[Themes] Upgrade to rtd 1.0.0rc1 and override incompatible default changes

wouterj:
This new version is a bit nicer imho :) (especially with definition lists)

We also changed some templates in the `default` theme for the symfony.com styling, but this broke the rtd theme. I've added the original templates to the rtd theme in order to fix our CI builds. I've also recommended using this tool to check transitions to reStructuredText for some bundle authors, so not seeing headlines and code examples is a bit of a shock for them.


wouterj:
@weaverryan @Nyholm is it possible to get this PR merged and tagged with a version? This only changes the "rtd" theme, which shouldn't affect symfony.com build pipelines.
This tool is _the_ way to let maintainers of third party bundles check their rst doc files, but we currently broke it :confused:

Nyholm (approved):

weaverryan:
Thanks Wouter!

PR #114, opened 2021-09-13
Fix some edge-case related to TOCs of docs without titles

javiereguiluz:
In some symfony.com docs we have some partial docs that are included in other docs ... but instead of naming them as `foo.rst.inc` as in Symfony Docs, we name them as `foo.rst`. This included file is just a paragraph without any titles, so when trying to generate the TOC it fails.


weaverryan (approved):
Looks good! Though, I think naming these "partials" with a `.rst` ending is weird - they would, I believe, be seen as "orphaned" documents in Sphinx.

wouterj:
Yeah, this should be named `.rst.inc` imho

PR #123, opened 2021-10-15
Add a failing test about tables

javiereguiluz:
The table with the leading space in the header and/or rows is failing.


weaverryan:
Fixed at https://github.com/doctrine/rst-parser/pull/165

PR #124, opened 2021-10-20
Add a failing test about diff code highlighting

javiereguiluz:
@wouterj reported that a code block here https://symfony.com/doc/current/security.html#form-login didn't show any highlighting.

It's true. The problem occurs when a `diff` code block doesn't start its first line with `+` or `-` In this failing test, the first code block works but the second doesn't ~(to be completely honest, the first one doesn't work either because of some white space issue which I can't see, but it mostly works)~ <-- this problem was fixed by Ryan. Thanks!


weaverryan:
Fixed at: https://github.com/doctrine/rst-parser/pull/168

PR #128, opened 2021-10-25
Add a failing test about email links

javiereguiluz:
If you look at the first paragraph of this section: https://symfony.com/doc/current/contributing/code_of_conduct/care_team.html#members you'll see an error with some bold text and an email link.

In this failing test, all cases work except `**user@example.com**` and `*user@example.com*`

If it's complicated to make it work, let's reword the docs to not use that syntax. Thanks!

### DIFF
--- a/tests/IntegrationTest.php
+++ b/tests/IntegrationTest.php
@@ -136,6 +136,10 @@ if(str_contains($expectedFile, 'datetime.html')) {
             'blockName' => 'nodes/list',
         ];
 
+        yield 'email' => [
+            'blockName' => 'nodes/email',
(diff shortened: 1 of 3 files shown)


stof:
Looking into https://github.com/doctrine/rst-parser/, it looks like the detection of standalone emails is done during the parsing phase, while the detection of emphasis (and other inline syntaxes) is done at render time. This might be the cause of the issue.

javiereguiluz:
I'm closing this issue here and moving it to doctrine/rst-parser to see if it can be fixed there. Thanks.

PR #131, opened 2021-11-04
Expose `console` as a real Composer binary

wouterj:
...and rename it to `docs-builder`


weaverryan (approved):
Fine with me: we clearly state this is an internal tool, so I don't have any problem with exposing a binary. Probably the symfony.com build process will just need to update after we tag this - so FYI @javiereguiluz

javiereguiluz:
I think this won't affect symfony.com because we don't use this tool via a high-level command run via Symfony's `Process` component (we did that in the past). Instead, we use it via low-level function calls such as:

```php
$buildConfig = (new BuildConfig())->...();
$result = (new DocBuilder())->build($buildConfig);
if ($result->isSuccessful()) { ... } else { ... }
```

This is the same I'm proposing in the "doc build instructions" at https://github.com/symfony/symfony-docs/pull/16068 Please, tell me what do you think about it. Thanks!

javiereguiluz:
Thank you Wouter.

PR #132, opened 2021-11-13
Improve copy/paste of diff code blocks

HypeMC:
Currently, copy-pasting code from diff blocks also copies the `+` & `-` signs, which can be really annoying. The idea here is to remove those signs from the html & add them via css & the `::before` pseudo element. This way the signs don't get copied with the rest of the code.

screen-copy


HypeMC:
Not need any more now that there's a before & after.

PR #133, opened 2021-11-18
Include all required options for a successful symfony-docs build

wouterj:
This modifies the script so that it can be run without any customization in the symfony-docs repository.

Having all PHP code of the docs builder in 1 repository means that we can more easily update the script in case of BC breaks (which will happen, as this package is internal).


javiereguiluz:
Thanks Wouter!!  About the errors reported by GitHub Actions ... are they related to these changes? They are a bit strange 😐

PR #136, opened 2021-12-24
Use github.com instead of api.symfony.com

wouterj:
ref https://github.com/symfony/symfony-docs/issues/16322


fabpot (approved):

OskarStark (approved):

javiereguiluz:
Thank you Wouter.

src/Reference/MethodReference.php
```diff
@@ -42,7 +42,7 @@ public function resolve(Environment $environment, string $data): ResolvedReferen
         return new ResolvedReference(
             $environment->getCurrentFileName(),
             $methodName.'()',
-            sprintf('%s/%s.html#method_%s', $this->symfonyApiUrl, str_replace('\\', '/', $className), $methodName),
+            sprintf('%s/%s.php#method_%s', $this->githubUrl, str_replace('\\', '/', $className), $methodName),
```
stof, line 45: these is no `method_...` id on GitHub.

wouterj: I know, here is me hoping they'll add this one day :roll_eyes:

OskarStark: I think they did sth. else for other languages already 🤔

stof: Well, their own code navigation logic is based on server-side knowledge of the lines are which you need to navigate. The generated links are using their `L*` anchors to navigate to specific line numbers. There is no special anchors in the page for the methods.

wouterj: If I'm correct, @OskarStark suggested them to add named anchors (i.e. class names or method names) besides line nr anchors and they were quite positive about that idea.

stof: Well, they haven't done it yet, then. And we don't know which naming pattern they will use if they finally do it.

OskarStark: Yes I did, but it's long time ago...

PR #149, opened 2023-02-14
Replace `{version}` in URLs

alamirault:
In order to make link on symfony codebase, even if is not a class or a method reference

Can be usefull in this case for example https://github.com/symfony/symfony-docs/pull/17902

### DIFF
--- a/src/KernelFactory.php
+++ b/src/KernelFactory.php
@@ -44,7 +44,8 @@ final class KernelFactory
             new SymfonyHTMLFormat(
                 $configuration->getTemplateRenderer(),
                 $configuration->getFormat(),
-                $urlChecker
+                $urlChecker,
+                $buildConfig->getSymfonyVersion()
             )
         );
 
--- a/src/Renderers/SpanNodeRenderer.php
+++ b/src/Renderers/SpanNodeRenderer.php
@@ -25,23 +25,27 @@ class SpanNodeRenderer extends AbstractSpanNodeRenderer
     private $decoratedSpanNodeRenderer;
     /** @var UrlChecker|null */
     private $urlChecker;
+    private $symfonyVersion;
 
     public function __construct(
         Environment $environment,
         SpanNode $span,
         BaseSpanNodeRenderer $decoratedSpanNodeRenderer,
-        ?UrlChecker $urlChecker = null
-    ) {
+        ?UrlChecker $urlChecker = null,
+        string $symfonyVersion = null
+    )
+    {
         parent::__construct($environment, $span);
 
         $this->decoratedSpanNodeRenderer = $decoratedSpanNodeRenderer;
         $this->urlChecker = $urlChecker;
+        $this->symfonyVersion = $symfonyVersion;
     }
 
     /** @inheritDoc */
     public function link(?string $url, string $title, array $attributes = []): string
     {
-        $url = (string)$url;
+        $url = (string) $url;
 
         if (
             $this->urlChecker &&
@@ -53,6 +57,10 @@ class SpanNodeRenderer extends AbstractSpanNodeRenderer
 
         if (!$this->isSafeUrl($url)) {
             $attributes = $this->addAttributesForUnsafeUrl($attributes);
+        }
+
+        if (null !== $this->symfonyVersion) {
+            $url = u($url)->replace('{version}', $this->symfonyVersion)->toString();
         }
 
         return $this->decoratedSpanNodeRenderer->link($url, $title, $attributes);
--- a/src/SymfonyHTMLFormat.php
+++ b/src/SymfonyHTMLFormat.php
@@ -29,12 +29,14 @@ final class SymfonyHTMLFormat implements Format
     private $htmlFormat;
     /** @var UrlChecker|null */
     private $urlChecker;
+    private $symfonyVersion;
 
-    public function __construct(TemplateRenderer $templateRenderer, Format $HTMLFormat, ?UrlChecker $urlChecker = null)
+    public function __construct(TemplateRenderer $templateRenderer, Format $HTMLFormat, ?UrlChecker $urlChecker = null, string $symfonyVersion = null)
     {
         $this->templateRenderer = $templateRenderer;
         $this->htmlFormat = $HTMLFormat;
         $this->urlChecker = $urlChecker;
+        $this->symfonyVersion = $symfonyVersion;
     }
 
     public function getFileExtension(): string
@@ -69,7 +71,8 @@ final class SymfonyHTMLFormat implements Format
                     $node->getEnvironment(),
                     $node,
                     new BaseSpanNodeRenderer($node->getEnvironment(), $node, $this->templateRenderer),
-                    $this->urlChecker
+                    $this->urlChecker,
+                    $this->symfonyVersion
                 );
             }
         );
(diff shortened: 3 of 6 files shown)


tests/fixtures/source/blocks/references/file.rst
```diff
@@ -0,0 +1,2 @@
+
+:file:`Link name::Symfony/Component/Notifier/Bridge/AmazonSns/README.md`
```
alamirault, line 2: Are you agree with this syntax ?

Do you have a better idea ?

OskarStark: As we link to GitHub code, I propose to have sth like:

`:file:symfony/symfony::src/......`

so we could also link to: `symfony/symfony-docs` repo too.

Maybe we should rename it anyway to `:github-file:` ?

alamirault: We want inject `symfonyVersion` for `symfony` and `symfony-docs` not other repository.

I suggest

``:github-file:`My Link::symfony::src/` `` produce `https://github.com/symfony/symfony/blob/6.3/src`

``:github-file:`My Link::symfony-docs::src/` `` produce `https://github.com/symfony/symfony-docs/blob/6.3/src`

``:github-file:`My Link::foo/bar::src/` `` produce `https://github.com/foo/bar/src`

This way, it's easy to configure for usual links and we can also configura any github paths

WDYT ?

OskarStark: If others agree I like it 😃

javiereguiluz: If we follow the traditional RST syntax:

```rst
:ref:`The Label <the-reference-id>`
:doc:`The Label </the/doc/path>`
```

This could be instead:

```rst
:github:`My Link <symfony/symfony:src/>`
:github:`My Link <symfony/symfony-docs:src/>`
:github:`My Link <foo/bar:src/>`
```

Comments:

* `:github-file:` could be shortened as `:github:`. If in the future we add support for linking to issues or PR, we could create `:github-issue:` but keep the short `:github:` for the most common case.
* I wouldn't add a special case for the `symfony/*` repos because that always increases the learning curve: What repo is `symfony` ?¡can't happen if we use the explicit `symfony/symfony`.
* About the separator between repo name and path, we could use just `:` instead of `::` ... because `:` can't happen in repo names or file paths.

In any case, I'm still not 100% sure about this. I know we're trying to solve a real problem we have, but creating many custom RST tags would create a meta-language based on RST, which complicates maintenance and docs contributions.

But let's hear the opinions from @xabbuh, @wouterj and @weaverryan. Thanks!

OskarStark: I understand your Problem, but that's exactly where the docs-builder can ease work.

I like and would go with:
```rst
:github:`My Link <symfony/symfony:src/>`
```

alamirault: I pushed a new version. If we don't want magic, branch must be explicit, we can use `{symfonyVersion}` special placeholder for symfony repos.

```rst
:github:`First link ::symfony/symfony:4.0/src/Symfony/Component/Notifier/Bridge/AmazonSns/README.md::`

:github:`Link with branch substitution ::foo/bar:{symfonyVersion}/src/baz::`
```
It seems complicated to use following syntax (I love it !)
```rst
`:github:`Link with branch substitution <foo/bar:{symfonyVersion}/src/baz>`
```
Indeed `rst-parser` consider it as a real `Doctrine\RST\Span\SpanToken` and value sent to our code (`$spanToken->get('url')`) is only `foo/bar:{symfonyVersion}/src/baz` (without link name) not `Link with branch substitution <foo/bar:{symfonyVersion}/src/baz>`

javiereguiluz: The other day we merged https://github.com/symfony/symfony-docs/pull/17967 and seeing these "URI templates", I come up with an alternative idea to solve this problem:

(1) What we want to do: link to some GitHub file, but use a different version for each doc branch:

https://github.com/symfony/amazon-sns-notifier/blob/5.4/README.md
https://github.com/symfony/amazon-sns-notifier/blob/6.1/README.md
https://github.com/symfony/amazon-sns-notifier/blob/6.2/README.md
https://github.com/symfony/amazon-sns-notifier/blob/6.3/README.md

(2) What if we use "URI templates" for this kind of links?

Instead of creating a new custom directive:

```rst
:github:`README file <symfony/amazon-sns-notifier:README.md/>`
:github:`README file <symfony/amazon-sns-notifier:{6.2}/README.md/>`
```

We could use the normal RST links but with variables inside the URL:

```rst
`README file`_

.. _`README file`: https://github.com/symfony/amazon-sns-notifier/blob/{version}/README.md

.. OR, embedded:

`README file <https://github.com/symfony/amazon-sns-notifier/blob/{version}/README.md>`_
```

-----

The advantages that I see to this approach is that:

* Docs are still very readable
* Maintenance is the same as before, not more complicated
* There's no need to create a custom directive

The disadvantages:

* We need to update `docs-builder` to process the links and replace `{version}` by the current Symfony version of the docs being parsed
* We need to be careful about the URI variable pattern; it needs to be `{version}`, not `{ version }`, etc.

What do you think?

OskarStark: Makes sense 👍 

> We need to be careful about the URI variable pattern; it needs to be {version}, not { version }, etc.

Can be ensured by DOCtor-RST 👍

alamirault: Thanks for your feedback. Native link with `{version}` substitution is the most readable indeed and easy to use.

I pushed new version in this direction.

Hope is ok :pray:

(I will create DOCtor-RST rule when this PR will be merge to ensure good usage)

javiereguiluz:
Antoine, thanks for updating the implementation of this PR.

I've reviewed it and it looks great to me 👍 

I only have a question ... I don't remember the details, but in symfony.com we build Symfony Docs and other docs different from Symfony (Cloud docs, bundles docs, etc.). So, what would happen if we use `{version}` in those docs? Do we have in this docs-builder app the concept of doc "version" instead of or in addition to "Symfony version"? Thanks!

alamirault:
Version is set with `BuildConfig`. I think each docs have specific value ?

https://github.com/symfony/symfony-docs/blob/43325d00d96ad51af2f90b61cb43e47b5918f450/_build/build.php#L22-L23

Maybe I have to rename `{version}` to `{symfonyVersion}` ?

Or introduce substitution config in BuildConfig, but I think it's overkill for now

javiereguiluz:
> Maybe I have to rename {version} to {symfonyVersion} ?

No ... my question was exactly about that ... if `{version}` is always "the version of the doc being rendered" ... that's perfect because this feature would work for ALL docs, and not ONLY for Symfony Docs.

alamirault:
> > Maybe I have to rename {version} to {symfonyVersion} ?
> 
> No ... my question was exactly about that ... if `{version}` is always "the version of the doc being rendered" ... that's perfect because this feature would work for ALL docs, and not ONLY for Symfony Docs.

It's always value of `->setSymfonyVersion()` setter

OskarStark:
> It's always value of ->setSymfonyVersion() setter

An that makes sense, the other docs should use absolute URLS.

@alamirault can you please rename the the PR title accordingly? I think it does not match the current PR content anymore

alamirault:
Title updated

javiereguiluz (approved):

OskarStark (approved):

OskarStark:
@weaverryan can you please merge and release this feature?

javiereguiluz:
Thanks Antoine for working on this feature, this is much appreciated.

PR #150, opened 2023-02-22
Made the configuration block tabs accessible

javiereguiluz:
This changes the generated HTML of `configuration-block` directive to make it fully accessible. I followed the https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/ recommendations that we also applied recently to Symfony Profiler.


javiereguiluz:
I'm merging this and tagging a new release. On symfony.com we've made the changes to support both the old and the new `configuration-block` HTML code.

PR #152, opened 2023-03-25
Avoid link with same ID's

alamirault:
Hello,

This PR try to generate unique link id.

Relative issues in symfony docs:

https://github.com/symfony/symfony-docs/issues/16417
https://github.com/symfony/symfony-docs/issues/16721

rst-parser instantiate NodeRenderer for each node so the only way I found is to use static property and save when id is already rendered on same page.

Examples of changes on symfony-docs:

https://github.com/alamirault/sf-doc-output/commit/45a528110ffd37982b1689b2b622a2a9edb40c9b

### DIFF
--- a/src/Renderers/TitleNodeRenderer.php
+++ b/src/Renderers/TitleNodeRenderer.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Docs Builder package.
+ * (c) Ryan Weaver <ryan@symfonycasts.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace SymfonyDocsBuilder\Renderers;
+
+use Doctrine\RST\Environment;
+use Doctrine\RST\Nodes\TitleNode;
+use Doctrine\RST\Renderers\NodeRenderer;
+use Doctrine\RST\Templates\TemplateRenderer;
+
+class TitleNodeRenderer implements NodeRenderer
+{
+    /** @var TitleNode */
+    private $titleNode;
+
+    /** @var TemplateRenderer */
+    private $templateRenderer;
+
+    private static $idUsagesCountByFilename = [];
+
+    public function __construct(TitleNode $titleNode, TemplateRenderer $templateRenderer)
+    {
+        $this->titleNode = $titleNode;
+        $this->templateRenderer = $templateRenderer;
+    }
+
+    public function render(): string
+    {
+        $filename = $this->titleNode->getEnvironment()->getCurrentFileName();
+        $id = $this->titleNode->getId();
+
+        $idUsagesCount = self::$idUsagesCountByFilename[$filename][$id] ?? 0;
+
+        if (0 === $idUsagesCount) {
+            $computedId = $this->titleNode->getId();
+        } else {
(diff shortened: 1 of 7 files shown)


src/Renderers/TitleNodeRenderer.php
```diff
@@ -0,0 +1,55 @@
...
+
+class TitleNodeRenderer implements NodeRenderer
+{
+    /** @var TitleNode */
+    private $titleNode;
+
+    /** @var TemplateRenderer */
+    private $templateRenderer;
+
+    private static $idRenderedByFilename = [];
```
alamirault, line 27: This way tests are broken, only when full suite is run. I don't know how I can fix it. (except a static method to reset array, and call it before running phpunit test)

src/Renderers/TitleNodeRenderer.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->titleNode = $titleNode;
+        $this->templateRenderer = $templateRenderer;
+    }
+
+    public function render(): string
+    {
+        $filename = $this->titleNode->getEnvironment()->getCurrentFileName();
+        $id = $this->titleNode->getId();
+
+        $actualCountRendered = self::$idRenderedByFilename[$filename][$id] ?? null;
```
javiereguiluz, line 40: I find this variable name a bit confusing. If I'm not the only one, here are some alternative suggestions: `$numTitleOccurrences`, `$idUsagesCount`, `$timesSameIdWasUsed`, etc.

Also, instead of `null` we could initialize it to `0`, to the variable is always of type `int`. Thanks

src/Renderers/TitleNodeRenderer.php
javiereguiluz: This is always tricky. I'd prefer to not add a method to the class if it's only used in tests.

We could use reflection in the tests:

```php
$reflection = new \ReflectionClass(TitleNodeRenderer::class);
$property = $reflection->getProperty('idRenderedByFilename');
$property->setAccessible(true);
```

src/Renderers/TitleNodeRenderer.php
alamirault: I agree, I updated to `idUsagesCount`

javiereguiluz (approved):
Thanks Antoine for fixing this annoying bug 🙇

alamirault:
New issue opened https://github.com/symfony/symfony-docs/issues/18194

Merging this PR will be very cool :smiley:

src/Renderers/TitleNodeRenderer.php
```diff
@@ -0,0 +1,55 @@
...
+
+use Doctrine\RST\Environment;
+use Doctrine\RST\Nodes\TitleNode;
+use Doctrine\RST\Renderers\NodeRenderer;
+use Doctrine\RST\Templates\TemplateRenderer;
+
+class TitleNodeRenderer implements NodeRenderer
+{
+    /** @var TitleNode */
+    private $titleNode;
```
OskarStark, line 22: @javiereguiluz can we use typed properties?

javiereguiluz: Let's check that ... but do it in other PR so we can merge and deploy this one as soon as possible. Thanks.

javiereguiluz:
Thanks Antoine! Let's now tag a new version and deploy it in production so these issues can be fixed soon. Cheers!

PR #154, opened 2023-04-03
Fix language labels array of ConfigurationBlockDirective

alexander-schranz:
Related to https://github.com/symfony-tools/docs-builder/pull/153 @javiereguiluz


javiereguiluz:
Thanks for fixing this bug Alexander.

PR #156, opened 2023-04-13
Fix the TOC generation with unique links

### LINKED ISSUE #155: Issue with the feature that avoids links with the same ID
The #152 fix works beautifully ... but there's something missing in it. The TOC links seems to be not updated.

Example:

* in the TOC: https://symfony.com/doc/current/components/string.html#slugger
* in the body: https://symfony.com/doc/current/components/string.html#slugger-1

Thanks

javiereguiluz:
Still WIP. It fixes #155.

I can't find any way of getting the `id` attributes of headings without changing the vendor dependencies ... so this proposes to just parse the generated HTML file and extract the information from it.

### DIFF
--- a/src/DocsKernel.php
+++ b/src/DocsKernel.php
@@ -20,6 +20,7 @@ use Doctrine\RST\Kernel;
 use SymfonyDocsBuilder\Listener\AdmonitionListener;
 use SymfonyDocsBuilder\Listener\AssetsCopyListener;
 use SymfonyDocsBuilder\Listener\CopyImagesListener;
+use SymfonyDocsBuilder\Listener\DuplicatedHeaderIdListener;
 
 class DocsKernel extends Kernel
 {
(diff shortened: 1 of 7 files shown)


javiereguiluz:
This is now ready for review.

@wouterj, @alamirault, @oskarstark if you have some time, please help me reviewing this. Thanks a lot!

javiereguiluz:
After the last changes, I rebuilt Symfony Docs locally and it worked as expected.

The problem was that the static variable that holds header IDs (which we need to be static for it to work) wasn't reset before parsing each document.

In practice, the code made header IDs unique across all files ... and now they are unique within each single file.

OskarStark:
Looks good to me

wouterj:
> In practice, the code made header IDs unique across all files ... and now they are unique within each single file.

Note that per spec, explicit references (the `.. _some-ref:` syntax) must be unique across all files whereas implicit references (the one automatically generated based on section titles) must be unique within each single file.

PR #161, opened 2023-06-07
Adding the "tabs" directive

weaverryan:
Hi!

This adds a "tabs" directive, based off of https://sphinx-tabs.readthedocs.io/en/latest/

Example:

```rst

.. tabs:: UX Installation

    .. tab:: Webpack Encore

        Webpack Encore stuff!

        .. code-block:: yaml

            I am yaml:

    .. tab:: AssetMapper

        AssetMapper stuff!

        And another paragraph.
```

This will render (with a little assistance from Javier) just like a "configuration block", except that you can put any content inside. My use-case is for some UX-related items where install steps will now be different if you're using Encore vs AssetMapper, and I think a tabbed setup will be better than 2 sections right after each other.

Feedback appreciated!


javiereguiluz:
@weaverryan thanks for this! I've tested this in my local machine and didn't fully work. The missing part is the `data-language` attribute for each tab with the same content as the tab title ("Webpack Encore" and "AssetMapper" in this case).

I know that these tabs won't always be used for "programming languages" ... but this is just an internal detail in the generated HTML, so I think it's OK to do it without renaming the `data-language` attribute. Thanks!

javiereguiluz:
Thanks for the updates! I copied the HTML code of `tests/fixtures/expected/blocks/directives/tabs.html` in a random Symfony Docs page and it worked:

tabs

About the value of `data-language`, it can be anything as long as it's different from the other tabs and it's consistent (all tabs of the same type use the same value).

So, this should be ready! Cheers!

weaverryan:
Yaaaaaay! That looks like I expected. Thank you!!!

weaverryan:
@javiereguiluz Is this ready to merge? I was going to allow you to do it, as I assume you'll be the one integrating this into symfony.com.

Thanks!

javiereguiluz:
Ryan, I'm really sorry! I totally forgot about this 😞  I'm merging this, tagging a new release and then updating symfony.com server to use the new version. Thanks!

javiereguiluz:
@weaverryan this is now deployed on symfony.com, we can already use the new directive in docs. Thanks!

weaverryan:
thank you!!!

PR #170, opened 2024-01-25
Test this project on more platforms

javiereguiluz:


javiereguiluz:
OK then, let's remove Windows and macOS and let's only update the PHP versions. Thanks.

PR #172, opened 2024-05-15
Highlight PHP attributes

javiereguiluz:
Our current highlighter doesn't support PHP attributes ... but we use more and more attributes in the Symfony Docs.

I tried changing the highlighter (I tested https://github.com/tempestphp/highlight) but, although it solved the attribute issue, they introduced more issues.

So, this PR adds a "hack" to the current highlighter to add support for PHP attributes:

**Before // After**

-----

I added many tests to ensure that we support all the different ways of defining PHP attributes. There's still an edge case related to multi-line attributes. It's not a big problem and I'll tackle this in a future PR.

### DIFF
--- a/src/Renderers/CodeNodeRenderer.php
+++ b/src/Renderers/CodeNodeRenderer.php
@@ -70,9 +70,10 @@ class CodeNodeRenderer implements NodeRenderer
 
             $highLighter = new Highlighter();
             $highlightedCode = $highLighter->highlight($languageMapping, $code)->value;
+        }
 
-            // this allows to highlight the $ in PHP variable names
-            $highlightedCode = str_replace('<span class="hljs-variable">$', '<span class="hljs-variable"><span class="hljs-variable-other-marker">$</span>', $highlightedCode);
+        if ('php' === $language) {
+            $highlightedCode = $this->processHighlightedPhpCode($highlightedCode);
         }
 
         if ('terminal' === $language) {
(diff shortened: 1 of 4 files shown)


wouterj:
I personally feel like we can better update the syntax file: https://github.com/symfony-tools/docs-builder/blob/main/src/Templates/highlight.php/php.json (which we already maintained to support newer syntax) than processing the resulting HTML output.

The latest highlight.js PHP syntax file also [supports PHP attributes already](https://highlightjs.org/demo#lang=php&v=1&theme=atom-one-dark&code=PD9waHAKCiAgICBuYW1lc3BhY2UgQXBwXFNvbWVQYXRoO8UcdXNlIFN5bWZvbnlcQ29tcG9uZW50XFZhbGlkYXRvclxDb25zdHJhaW50cyBhcyBBc3NlcnTGO2NsYXNzIMRPQ8QKxRR7xQbEASNbQXR0cmlidXRlTmFtZV3JGXByaXZhdGUgJHByb3BlcnR5McZP0zUoKds3Mto3J3ZhbHVlJ9w%2BM98%2BZScsIG9wdGlvbjog30%2FETzTaT1vHPyA9PshKXdxLNdpLyRnEAcdMLM0V7wCzyRzccTbNceUB4Vz%2FAVDSRTffRf8Avf8Avdp4OMx4Um91dGUoJy9ibG9nL3twYWdlPFxkKz59JyzlAwQ6ICfEGl9saXN0%2FQDNOfMAzUdyZWF0ZXJUaGFuT3JFcXVhbO4A0uUAtDogMTjqANTcYzEwzGRPUk1cQ3VzdG9tSWRHZW5lcuQDfCjlA2E6ICdkb2N0cmluZS51dWlkX2fIIP0AwzHnA1Z9), so we can take inspiration from it. Unfortunately, I've tried using the converter from HighlightPHP in the past but this no longer works with the latest highlight.js syntax files so we can't automate it :(

javiereguiluz:
Closing in favor of #173, which does the same but it's much simpler and easier to maintain.

PR #173, opened 2024-05-17
Update the php syntax file to highlight attributes

wouterj:
Alternative to https://github.com/symfony-tools/docs-builder/pull/172 (as per my suggestion).

@javiereguiluz what do you think of this? The output is slightly different from yours, but that's how highlight.js works :)

I tried my suggestion of converting the latest highlight.js syntax file, this unfortunately didn't work (there are too many BC breaks in v10). So I've copy pasted their regex and used it to what is available in our syntax file already.

If we want to go this way, I can also follow up with a PR adding more support for PHP features (e.g. arrow functions). For now, I've just added a couple keywords that plainly were missing (such as trait and enum).

### DIFF
--- a/src/Templates/highlight.php/php.json
+++ b/src/Templates/highlight.php/php.json
@@ -8,8 +8,23 @@
         "php7"
     ],
     "case_insensitive": true,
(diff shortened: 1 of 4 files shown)


javiereguiluz:
@wouterj this is beautiful 🙌 ... and so simple compared to #172 😍 

Thanks a lot for working on this. I just closed the other PR.

I'm going to merge this and tag a new version so we can test this in symfony.com docs as soon as possible. Thanks!

javiereguiluz:
@smnandre thanks! I did a POC using tempest ... it highlights PHP attributes, but it uses a highlighting code that didn't play well with other existing highlighting. Also, we need to highlight C, Markdown, etc. for some parts of Symfony Docs and those are not supported (yet) in tempest.

So,we're putting that idea on hold for now ... but we'll see in the future. Your experience using it in ux.symfony.com will be vey valuable when we revisit this idea. Thanks!

PR #175, opened 2024-05-21
Fixes for PHP attribute parsing

### LINKED ISSUE #174: Minor issue when highlighting some PHP attributes
The PHP attribute highlight feature that @wouterj implemented in #173 works really well. However, I found an edge case. See:

https://symfony.com/doc/current/reference/constraints/AtLeastOneOf.html#basic-usage

The `<span>` is applied to the entire PHP attribute, so everything is highlighted as an attribute in this case even if statements like `new Assert\Regex('/#/')` should be highlighted differently.

In any case, this is very rarely used in Symfony Docs, so it's fine to keep this edge case unfixed if the possible solution is very complex or breaks other things. Thanks!

wouterj:
- No longer wrap the entire attribute in a span, but only wrap `#[Route` and `]` in spans (fixes #174)
- @javiereguiluz for consistency with the real PHP highlight.js file, I've renamed `php-attributes` to `meta`. Is it OK for you to update the CSS on symfony.com to style `.hljs-meta` instead of `.hljs-php-attributes`?
- Fixed not parsing attributes in parameters, see e.g. https://symfony.com/doc/current/controller.html#mapping-the-whole-query-string

### DIFF
--- a/src/Templates/highlight.php/php.json
+++ b/src/Templates/highlight.php/php.json
@@ -11,17 +11,30 @@
     "keywords": "and include_once list abstract global private echo interface as static endswitch array null if endwhile or const for endforeach self var while isset public protected exit foreach throw elseif include __FILE__ empty require_once do xor return parent clone use __CLASS__ __LINE__ else break print eval new catch __METHOD__ case exception default die require __FUNCTION__ enddeclare final try match switch continue endfor endif declare unset true false goto instanceof insteadof __DIR__ __NAMESPACE__ yield finally",
     "contains": [
         {
-            "className": "php-attribute",
-            "begin": "#\\[\\s*(\\\\?[A-Z][A-Za-z0-9_\\x7f-\\xff]+|\\\\?[A-Z]+(?=[A-Z][a-z0-9_\\x7f-\\xff])){1,}(?![A-Za-z0-9])(?![$])",
+            "className": "meta",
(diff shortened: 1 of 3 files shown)


wouterj:
Looking at the tests, this unfortunately this apparently broke other PHP syntax. Needs more investigation.

wouterj:
This PR is ready to merge now. Fix was to reorder the syntax elements in the parameter list, no clue why but it seems to work (also checked by building the full docs locally).

> I've just updated symfony.com to add this: `.hljs-php-attribute, .codeblock-php .hljs-meta { ... }` and I'll remove `.hljs-php-attribute` when the docs are rebuilt with this new version. So, everything is ready on that side.

You might want to make it `.hljs-meta` (so without scoping to only PHP codeblocks). Meta is used for special meta syntax in other formats (e.g. the doctype in HTML/XML, `use strict` in JavaScript and anchors/references in Yaml).

javiereguiluz:
Great! Wouter, thanks again for this super nice contribution. I'm going to merge and tag a new release to use this on symfony.com ASAP.

javiereguiluz:
After deploying and rebuilding the docs, the highlighting of PHP attributes is now perfect. See e.g. the page that was showing the issue before: https://symfony.com/doc/current/reference/constraints/AtLeastOneOf.html#basic-usage

PR #176, opened 2024-05-27
Add more modern PHP syntax highlighting

wouterj:
Because I had fun playing around with this, I've added more changes to the PHP syntax file, bringing it more or less up to date with Highlight.js' syntax file.

Noteworthy changes:
* Support arrow functions, enums, match statement, named arguments and more constants
* Updated the full list of supported PHP types
* Identify function class/class instantiations (this improves highlighting function arguments)

@javiereguiluz please test this out locally. This patch will generally add more color to the code blocks. You might want to tweak the CSS a bit depending on your favor (we seem to have relatively little coloring in our code examples).
E.g. function calls now get `hljs-title` as well (you can use `.invoke__` to differentiate them from other titles), and named arguments will be labelled as `hljs-attr` (which is currently colored similar to `hljs-meta` used for PHP attributes).

| Before | After
| --- | ---
| image | image
| image | image
| image | image

### DIFF
--- a/src/Renderers/CodeNodeRenderer.php
+++ b/src/Renderers/CodeNodeRenderer.php
@@ -70,9 +70,6 @@ class CodeNodeRenderer implements NodeRenderer
 
             $highLighter = new Highlighter();
             $highlightedCode = $highLighter->highlight($languageMapping, $code)->value;
-
-            // this allows to highlight the $ in PHP variable names
-            $highlightedCode = str_replace('<span class="hljs-variable">$', '<span class="hljs-variable"><span class="hljs-variable-other-marker">$</span>', $highlightedCode);
         }
 
         if ('terminal' === $language) {
--- a/src/Templates/highlight.php/php.json
+++ b/src/Templates/highlight.php/php.json
@@ -8,7 +8,7 @@
         "php7"
     ],
     "case_insensitive": true,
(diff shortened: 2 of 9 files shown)


wouterj:
@javiereguiluz I just rebased this PR and made it green. So it's ready for you to try out locally whenever you want :) (no pressure at all, things won't break)

OskarStark:
Wo, I love it, thanks @wouterj !

javiereguiluz:
Wouter, thanks a lot for this great work. What a nice addition to the PHP highlighting 🙇🙇

I've been playing with it locally and it works exactly as expected:

* It improves many edge cases that previously weren't highlighted...
* ... without breaking any of the existing highlighting

A before/after example using one of the same examples that you showed:

image

-----

So, we can totally merge this and start using it in production. Thanks!

wouterj:
Thanks for testing Javier!

I'll let you handle releasing a version as you can best coordinate this with the website :)

javiereguiluz:
Thanks! This was released as `v0.26.0` in https://github.com/symfony-tools/docs-builder/releases/tag/v0.26.0

stof:
As those PHP highlighting rules have been forked from the ones in the highlight.php library, which were based on the highlight.js project, it would be great to identify which improvements could be submitted to upstream projects. Maybe at some point in the future, we would be able to avoid maintaining custom highlighting rules for PHP and Twig.

wouterj:
@stof in an ideal world, yes. But in practical world, the reality is that the PHP fork is more or less dead. The project relies on Highlight.js's JS syntax files and has a script to convert them to JSON. However, since Highlight.js 10 the JS syntax files can contain functions that will be used when highlighting code. These of course can not be converted to JSON, which broke the convert script.

There is an experiment to create a JS code to PHP code converter (https://github.com/scrivo/highlight.php/pull/109), but that seems rather stalled (and sounds fragile). So we are forever stuck to the syntax files of v9.
What I did in essence in this PR is manually convert the newer PHP syntax file to our file.

On the long term, we probably would be better of to use another highlighting library. In the mean time, we can keep manually updating the PHP syntax file (given that is by far the most significant language in the Symfony docs).

PR #178, opened 2024-07-31
Fix attribute arguments highlighting and better support CPP

wouterj:
@javiereguiluz after seeing my changes online, I discovered 2 minor bugs that are fixed with this PR:
- Highlight `public`/`protected`/`private` when using constructor property promotion
- Fix highlighting attribute arguments (these were now parsed as function definition parameters instead of invocation arguments, e.g. lacking named argument support)


javiereguiluz:
This is great! Wouter, you keep improving highlighting more and more 🙌  Thanks 🙇

PR #179, opened 2024-08-01
:bug: Fix yaml number rendering

### LINKED ISSUE #177: Minor YAML highlighting issue
First reported here: https://github.com/symfony/symfony-docs/issues/20078

When using underscores to make config more readable, some numeric values are wrongly (and partially) highlighted as strings instead of numbers:

image

homersimpsons:
Fixes #177 

I copied the yaml.json file from https://github.com/scrivo/highlight.php/blob/d1c6b0956a2b0d3efc137e4d8c5bf5c5e220bac1/src/Highlight/languages/yaml.json#L84 and updated the "number" regex to add `_` where I think the symfony parser allow them:
```diff
- "begin": "(-?)(\\b0[xX][a-fA-F0-9]+|(\\b\\d+(\\.\\d*)?|\\.\\d+)([eE][-+]?\\d+)?)\\b"
+ "begin": "(-?)(\\b0[xX][a-fA-F0-9_]+|(\\b\\d[\\d_]*(\\.[\\d_]*)?|\\.\\d[\\d_]*)([eE][-+]?\\d[\\d_]*)?)\\b"
```

There may be additional places where the symfony parser allow the `_` for numbers.


wouterj (approved):
Thanks for the fix!

Overwriting the syntax file in this project is indeed the most sensible option (see also my more lengthy response on this matter: https://github.com/symfony-tools/docs-builder/pull/176#issuecomment-2260301823).

The [highlight.js syntax file](https://github.com/highlightjs/highlight.js/blob/main/src/languages/yaml.js) has support for some more [number types](https://github.com/highlightjs/highlight.js/blob/main/src/languages/yaml.js#L79-L86) (especially timestamps, datetimes, etc.). Maybe we can also add those regexes (as a separate number item)? (given the Yaml component supports date strings)

homersimpsons:
> The [highlight.js syntax file](https://github.com/highlightjs/highlight.js/blob/main/src/languages/yaml.js) has support for some more [number types](https://github.com/highlightjs/highlight.js/blob/main/src/languages/yaml.js#L79-L86) (especially timestamps, datetimes, etc.). Maybe we can also add those regexes (as a separate number item)? (given the Yaml component supports date strings)

I can add those, but are they supported by `symfony/yaml`'s parser? I saw we were parsing dates but not sure about the exact syntax supported.

homersimpsons:
Just tested it. Their regex is pretty eager:
- It allows any year of 4 digit, I sticked with 2024 (I may have found a symfony bug without this constraint)
- It allows having a fractional time part without having a time
 

```php
var_dump(Yaml::parse(<<<YAML
v0: 2024-12-07 23:44:39.35677451141299410 	    	    		   		 	 	 	  	 	    		   			  		 			 			 				  	 		
v1: 2024-02-3.671	-11
v2: 2024-82T3:66:07		+36
v3: 2024-07-05t6:39:20	 	   		 	        	
v4: 2024-01-30.000				 	 	  	  				 							 	 	 	  	 	 			      			 				  	 	     	   	   	 		    	
v5: 2024-09-03 			 					  	   			    	   	  	    		  	
YAML
, Yaml::PARSE_DATETIME));
```

will give the following, we can see that symfony won't handle all possible "dates":

```
array(6) {
  'v0' =>
  class DateTimeImmutable#4 (3) {
    public $date =>
    string(26) "2024-12-07 23:44:39.356774"
    public $timezone_type =>
    int(3)
    public $timezone =>
    string(3) "UTC"
  }
  'v1' =>
  string(17) "2024-02-3.671     -11"
  'v2' =>
  string(20) "2024-82T3:66:07           +36"
  'v3' =>
  class DateTimeImmutable#5 (3) {
    public $date =>
    string(26) "2024-07-05 06:39:20.000000"
    public $timezone_type =>
    int(3)
    public $timezone =>
    string(3) "UTC"
  }
  'v4' =>
  string(14) "2024-01-30.000"
  'v5' =>
  class DateTimeImmutable#6 (3) {
    public $date =>
    string(26) "2024-09-03 00:00:00.000000"
    public $timezone_type =>
    int(3)
    public $timezone =>
    string(3) "UTC"
  }
}
```

@wouterj Should I still use their regex of find one that fits symfony/yaml better?

homersimpsons:
@wouterj I used the regex from [`symfony/yaml`](https://github.com/symfony/yaml/blob/fa34c77015aa6720469db7003567b9f772492bf2/Inline.php#L810-L826). I had to clean it up so it works in JS (removed the capture groups), and I also removed the non-capture groups that are unused.

javiereguiluz (approved):
Nice work! Thanks @homersimpsons

wouterj (approved):
Nice, this is looking great!

Failure is indeed not related to this PR.

javiereguiluz:
Merged! Thanks

PR #181, opened 2024-09-11
Add support for tilde characters

javiereguiluz:
Related to
* https://github.com/symfony/symfony-docs/issues/20215

This PR adds a failing test that shows the issue. This is the output change that would make test pass:

```diff
-<p>Excepteur &nbsp;1,000 sint! ...
+<p>Excepteur ~1,000 sint! ...
```

### DIFF
--- a/src/Renderers/SpanNodeRenderer.php
+++ b/src/Renderers/SpanNodeRenderer.php
@@ -42,6 +42,19 @@ class SpanNodeRenderer extends AbstractSpanNodeRenderer
         $this->symfonyVersion = $symfonyVersion;
     }
 
+    public function render(): string
+    {
+        // Work around "~" being parsed as non-breaking space by rst-parser,
+        // while this is not part of the specification.
+        $spanValue = $this->span->getValue();
+        $spanValue = str_replace('~', '__TILDE__', $spanValue);
+        $this->span->setValue($spanValue);
+
+        $rendered = parent::render();
+
+        return str_replace('__TILDE__', '~', $rendered);
+    }
+
     /** @inheritDoc */
     public function link(?string $url, string $title, array $attributes = []): string
     {
--- a/tests/IntegrationTest.php
+++ b/tests/IntegrationTest.php
@@ -142,6 +142,10 @@ class IntegrationTest extends AbstractIntegrationTest
             'blockName' => 'nodes/span-link',
         ];
 
+        yield 'text' => [
+            'blockName' => 'nodes/text',
+        ];
+
         yield 'title' => [
             'blockName' => 'nodes/title',
         ];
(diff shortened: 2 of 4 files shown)


stof:
doctrine/rst-parser parses `~` as a non-breaking space. this was introduced in https://github.com/doctrine/rst-parser/pull/42 but I have no idea where this comes from. I see no mention of non-breaking spaces in the reStructuredText documentation on the Sphinx website.

stof:
even worse, I see no way of *escaping* a `~` in the rst-parser library.

stof:
Maybe it would make sense to investigate migrating to `phpdocumentor/guides` as the implementation of the docs-builder

wouterj:
> Maybe it would make sense to investigate migrating to `phpdocumentor/guides` as the implementation of the docs-builder

We're already working on this (see https://github.com/wouterj/symfony-docs-guides), but more work is needed here (especially around ordering documents).

wouterj:
I've pushed a not-so-nice workaround to this PR. I don't see any other way to escape the tilde, and it should be safe given this is the Symfony docs specific builder.

src/Renderers/SpanNodeRenderer.php
```diff
@@ -42,6 +42,19 @@ public function __construct(
...
+    {
+        // Work around "~" being parsed as non-breaking space by rst-parser,
+        // while this is not part of the specification.
+        $spanValue = $this->span->getValue();
+        $spanValue = str_replace('~', '__TILDE__', $spanValue);
+        $this->span->setValue($spanValue);
+
+        $rendered = parent::render();
+
+        return str_replace('__TILDE__', '~', $rendered);
```
stof, line 55: Let's hope that we won't have content containing the actual text `__TILDE__` in the symfony documentation (or the documentation of other projects built with this builder).
Maybe we should check for `__TILDE__` in the spanValue before doing the `~` replacement and failing explicitly instead of producing a wrong output (or even smarter: selecting another non-conflicting placeholder in that case, but that might not be worth it)

javiereguiluz:
Thanks @wouterj for this contribution. Yes, I think this is a reasonable solution while we keep using the old RST parser. Thanks!

src/Renderers/SpanNodeRenderer.php
stof: Well, I would add least add the safety check to rebuild a build failure in case the assumption is not actually safe (instead of waiting for someone to discover that some obscure page has content not rendered like the source:

```php
if (str_contains($spanValue, '__TILDE__')) {
    throw new \Exception('Cannot render content containing the text "__TILDE__" as it is used as a special placeholder in the build.');
}
```

javiereguiluz: @stof OK, I've added your code while merging as an extra protection measure. Thanks!

PR #184, opened 2024-09-27
Add support for inline Twig comments

javiereguiluz:
Twig added inline comments (https://github.com/twigphp/Twig/pull/4349) so I tried to add support for them here. But, it's not working. Anyone knows if this could be easy to fix or if it's too complex to do it? Thanks!

### DIFF
--- a/src/Templates/highlight.php/twig.json
+++ b/src/Templates/highlight.php/twig.json
@@ -26,12 +26,21 @@
             "end": "%}",
             "contains": [
                 {
+                    "className": "comment",
+                    "begin": "#",
+                    "end": "$",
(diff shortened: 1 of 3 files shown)


wouterj:
We are green now :)

javiereguiluz:
Wouter, you are amazing! 🙇 Thanks a lot for fixing this.

javiereguiluz:
After merging and releasing this, I rebuilt the Symfony Docs and we can see that this is already working: https://symfony.com/bundles/ux-toggle-password/current/index.html#usage-without-symfony-forms

image

Thanks!

PR #195, opened 2025-11-05
Support multiple attributes in a single block

### LINKED ISSUE #194: Minor highlighting issue
This is just a rare edge-case, so we don't have to fix this. But, in case someone wants to contribute a fix, the current highlighter can't highlight two or more PHP attributes defined on the same line, as in:

```php
#[Argument, Ask('Enter the cloud region name')]
CloudRegion $region,
```

See https://symfony.com/blog/new-in-symfony-7-4-improved-invokable-commands#:~:text=If%20your%20interactions%20only%20involve%20asking%20for%20missing%20argument%20or%20option%20values%2C%20you%20can%20make%20them%20even%20more

wouterj:
Fixes #194 

@javiereguiluz this contains one change for all attribute highlighting: the whole `#[...]` section is now wrapped in a single `.hljs-meta` element. This means that everything without style now also gets the green text color on symfony.com (e.g. parenthesis and commas).

### DIFF
--- a/src/Templates/highlight.php/php.json
+++ b/src/Templates/highlight.php/php.json
@@ -12,30 +12,29 @@
     "contains": [
         {
             "className": "meta",
-            "begin": "#\\[\\s*(\\\\?[A-Z][A-Za-z0-9_\\x7f-\\xff]+)+\\]"
-        },
-        {
-            "begin": "#\\[\\s*(\\\\?[A-Z][A-Za-z0-9_\\x7f-\\xff]+)+(?![A-Za-z0-9])(?![$])",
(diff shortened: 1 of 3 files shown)


javiereguiluz:
This was released in this new tag:
https://github.com/symfony-tools/docs-builder/releases

And, it's already deployed in production and the blog post shows the new behavior:
https://symfony.com/blog/new-in-symfony-7-4-improved-invokable-commands#:~:text=If%20your%20interactions%20only%20involve%20asking%20for%20missing%20argument%20or%20option%20values

Thanks!

PR #196, opened 2025-11-17
Update the class reference directive to support Symfony AI classes

javiereguiluz:
We need to publish Symfony AI docs on symfony.com ASAP. The GitHub links generated for the classes used in Symfony AI docs are wrong, so we must fix that somehow. This PR proposes a possible fix.

In the future we'll need something better ... but we'r ealso migrating to another doc builder project, so it's not worth it to try to make it perfect now.


javiereguiluz:
I'm merging without waiting for reviews 🙏  ... but we can make new PRs later tof ix this in a different way if needed. Thanks!

PR #200, opened 2025-11-30
phpDocumentor Guides integration

wouterj:
This merges the SymfonyExtension from https://github.com/wouterj/symfony-docs-guides into this project.

Todo
---

* [x] Add fjson output support (https://github.com/wouterj/symfony-docs-guides/pull/8)
* [x] Add missing tabs directive (preferably, in phpDocumentor: https://github.com/phpDocumentor/guides/pull/1316)
* [x] Upgrade to PHP 8.5 and Symfony 8 (requires https://github.com/phpDocumentor/guides/pull/1317)
* [ ] Improve asset handling in phpDocumentor so we can store images and docs separately (symfony.com requirement). https://github.com/phpDocumentor/guides/pull/1362
* [x] Make a working prototype in the symfony.com repository, probably requires some updates to the `DocsKernel` and `DocBuilder` in this repository
* [x] Enable relative links, instead of absolute ones
* [ ] Report build warnings/errors as return value of `DocBuilder::build()`
* [ ] Implement missing fjson TOC information
* [ ] Cleanup this repository, removing Doctrine RST Parser related code. Also: discuss if we want to retain functionality like PDF generation, link checking.
* [ ] Restore license/copyright data where applicable


src/Application.php
```diff
@@ -1,42 +1,31 @@
 <?php
 
 /*
- * This file is part of the Docs Builder package.
- * (c) Ryan Weaver <ryan@symfonycasts.com>
+ * This file is part of the Guides SymfonyExtension package.
```
stof, line 4: This looks wrong to me.

src/TextRole/ClassRole.php
```diff
@@ -0,0 +1,50 @@
...
+        private BuildConfig $buildConfig,
+    ) {
+    }
+
+    #[\Override]
+    public function processNode(DocumentParserContext $documentParserContext, string $role, string $content, string $rawContent): InlineNodeInterface
+    {
+        $fqcn = u($content)->replace('\\\\', '\\');
+
+        $url = \sprintf($this->buildConfig->getSymfonyRepositoryUrl(), $fqcn->replace('\\', '/').'.php');
```
stof: This is loosing the changes done in recent PRs on this repo related to the support of the AI and UX monorepos.

resources/highlight.php/php.json
```diff
@@ -0,0 +1,275 @@
+{
+    "aliases": [
```
stof, line 2: Why are those highlight files new files instead of keeping (or moving) the existing files ?

composer.json
```diff
@@ -14,24 +14,42 @@
         }
     },
     "require": {
-        "php": ">=8.3",
+        "php": ">=8.2",
```
stof: Why adding support for older versions ?

src/Application.php
```diff
@@ -1,42 +1,31 @@
...
 
-    public function run(InputInterface $input): int
+    #[\Override]
+    public function run(?InputInterface $input = null, ?OutputInterface $output = null): int
     {
-        $inputOption = new InputOption(
-            'symfony-version',
-            null,
-            InputOption::VALUE_REQUIRED,
-            'The symfony version of the doc to parse.',
```
stof, line 34: Is this global CLI option defined elsewhere or is it gone ? Removing it might break the symfony.com website integration.

src/GuidesExtension/src/Build/StringBuildEnvironment.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    private FileSystem $filesystem;
+
+    public function __construct(string $contents)
+    {
+        $this->filesystem = new FlysystemV3(new LeagueFilesystem(new InMemoryFilesystemAdapter()));
+        $this->filesystem->put('index.rst', $contents);
+    }
+
+    #[\Override]
```
stof, line 29: we don't use this attribute in Symfony (and Psalm has a config setting to avoid requiring its usage)

composer.json
javiereguiluz: Note. On symfony.com we run 8.4 and soon, 8.5 so we can safely use modern PHP versions.

src/GuidesExtension/src/Application.php
```diff
@@ -1,42 +1,31 @@
...
+final class Application extends SymfonyApplication
 {
-    private $application;
-    private $buildConfig;
-
-    public function __construct(string $symfonyVersion)
-    {
-        $this->application = new BaseApplication();
-        $this->buildConfig = new BuildConfig();
+    public function __construct(
```
stof, line 20: This change breaks the `bin/docs-builder` script currently.

stof: @javiereguiluz is the symfony.com website using the `docs-builder` executable or is it using this package as a library ?

javiereguiluz: We use it as a library

wouterj:
Thanks for the detailed review, @stof :)

This was just a blunt import of the commits I made in my test project, which I started from scratch. So the goal now is to make sure the code in this PR matches whatever we were doing in this repository before. Your comments are a great start with this, as they show some unintentionally difference that we should revert from this PR before merging.

src/GuidesExtension/src/Application.php
stof: Could you share the integration code ? It would make it easier to figure out how to avoid (or at least reduce) BC breaks.

stof:
The CI also tells me that existing tests must be updated (which will help avoiding regressions if they can still run)

src/GuidesExtension/src/Application.php
javiereguiluz: For full docs (Symfony Docs, bundles, the Symfony Book) it's basically this:

```php
$buildConfig = (new BuildConfig())
    ->setContentDir($buildDir)
    ->setOutputDir($tempBuildDir)
    ->setImagesDir($publicImagesDir)
    ->setImagesPublicPrefix($publicImagesPrefix);

if ('symfony' === $bookSlug) {
    $buildConfig->setSymfonyVersion($branch);
}

try {
    $result = $this->docBuilder->build($buildConfig);

    if (!$result->isSuccessful()) {
        // ...
    }
} catch (\Throwable $e) {
    // ...
}
```

For blog posts, it's just this:

```php
$parseResult = (new DocBuilder())->buildString($rstContents);
if (!$parseResult->isSuccessful()) {
    // ...
}
```

stof: @wouterj this means it would be great to revert changes done to the DocBuilder public API if possible (I think we could expose the same public API than today based on top of guides) as it is the public API of this library.

Note that I don't even see where your new code uses the modified DocBuilder API (except in tests that haven't been updated and so are failing)

wouterj: Yes yes. I now also have access to the code behind the symfony.com's docs rendering. So I've started working on the integration (and making the new SymfonyExtension installable is a big part of that). Small steps, but we'll get there :)

wouterj:
Little update after this weekend: I've applied the changes applied to this repository since starting with the guides one 2 years ago (like AI github links and code highlight changes).

I've also merged the integration test fixtures from the Guides integration and this repository. This allows you to see the differences in the output, although 99% of it is whitespace so it's very hard to review. While working on the integration, I've been very careful with updating the expected files so I'm very confident the resulting HTML is still equal.

Next up: *moved to PR description*

tests/fixtures/expected/blocks/directives/caution.html
```diff
@@ -1,6 +1,7 @@
-<div class="admonition admonition-caution ">
+<div class="admonition admonition-caution">
     <p class="admonition-title">
-        <svg xmlns="http://www.w3.org/2000/svg" fill="none" width="24" height="24" viewBox="0 0 24 24" stroke="currentColor"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M12 9v2m0 4h.01m-6.938 4h13.856c1.54 0 2.502-1.667 1.732-3L13.732 4c-.77-1.333-2.694-1.333-3.464 0L3.34 16c-.77 1.333.192 3 1.732 3z" /></svg>
+        <svg xmlns="http://www.w3.org/2000/svg" fill="none" width="24" height="24" viewBox="0 0 24 24" stroke="currentColor"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M12 8v4m0 4h.01M21 12a9 9 0 11-18 0 9 9 0 0118 0z" /></svg>
```
wouterj: update caution icon

javiereguiluz:
Wouter, thanks a lot for the detailed update and for all this work!

src/GuidesExtension/src/DocBuilder.php
```diff
@@ -0,0 +1,75 @@
...
+use phpDocumentor\Guides\Handlers\ParseDirectoryCommand;
+use phpDocumentor\Guides\Handlers\RenderDocumentCommand;
+use phpDocumentor\Guides\Nodes\DocumentNode;
+use phpDocumentor\Guides\RenderContext;
+use phpDocumentor\Guides\Twig\Theme\ThemeManager;
+use SymfonyTools\DocsBuilder\GuidesExtension\Build\BuildConfig;
+use SymfonyTools\DocsBuilder\GuidesExtension\Build\BuildEnvironment;
+use SymfonyTools\DocsBuilder\GuidesExtension\Build\StringBuildEnvironment;
+
+final class DocBuilder
```
stof, line 26: shouldn't this DocBuilder class stay outside the `GuidesExtension` namespace ? This looks like the entrypoint *using* guides more than an extension of the guides library.

composer.json
```diff
@@ -5,33 +5,53 @@
     "license": "MIT",
     "autoload": {
         "psr-4": {
+            "SymfonyTools\\DocsBuilder\\GuidesExtension\\": "src/GuidesExtension/src",
             "SymfonyDocsBuilder\\": "src"
```
stof, line 9: is this `SymfonyDocsBuilder` namespace still used ?

composer.json
```diff
@@ -5,33 +5,53 @@
...
-        "symfony/http-client": "^5.2 || ^6.0 || ^7.0 || ^8.0",
-        "twig/twig": "^2.14 || ^3.3"
+        "phpdocumentor/guides": "^1.9",
+        "phpdocumentor/guides-cli": "^1.9",
+        "phpdocumentor/guides-code": "^1.7",
+        "phpdocumentor/guides-restructured-text": "^1.9",
+        "scrivo/highlight.php": "^9.12.0",
+        "symfony/config": "^6.0",
+        "symfony/filesystem": "^6.0",
+        "symfony/finder": "^6.0",
```
stof: shouldn't we allow newer Symfony versions ?

composer.json
```diff
@@ -5,33 +5,53 @@
...
-        "symfony/process": "^5.2 || ^6.0 || ^7.0 || ^8.0",
-        "masterminds/html5": "^2.7"
+        "league/flysystem-memory": "^3.0",
+        "symfony/phpunit-bridge": "^7.4@dev",
+        "symfony/process": "^6.0",
+        "symfony/var-dumper": "^6.2"
     },
-    "bin": ["bin/docs-builder"]
+    "scripts": {
+        "test": "SYMFONY_PHPUNIT_VERSION=12 simple-phpunit",
```
stof: I would suggest moving away from this `simple-phpunit` in favor of installing `phpunit/phpunit` as a dev requirement, which is the recommended approach for modern Symfony projects now.
It only involves registering the SymfonyExtension explicitly in the phpunit.xml file.

tests/fixtures/source/json/orphan.rst
```diff
@@ -1,3 +1,5 @@
+:orphan:
+
```
wouterj, line 2: ! orphaned documents are not automatically detected

(probably not a major blocker: symfony.com only renders Prev/Next page links for very specific sections of the docs)

src/GuidesExtension/src/Renderer/JsonRenderer.php
```diff
@@ -0,0 +1,79 @@
...
+
+            $prevDocument = $nextDocument = null;
+            if (!$documentNode->isOrphan()) {
+                $prevDocument = $context->getIterator()->previousNode();
+                $nextDocument = $context->getIterator()->nextNode();
+            }
+            $context->getDestination()->put(
+                $context->getDestinationPath().'/'.$context->getCurrentFileName().'.fjson',
+                json_encode([
+                    'parents' => [],
```
wouterj, line 56: No easy support for fetching the parent document. As far as I can find the symfony.com code, this property is never used so probably safe to ignore (cc @javiereguiluz please double check my assumption 😄)

resources/highlight.php/php.json
wouterj: they were copied, but the merge of 2 repositories messed it up. PR diff is corrected now :)

composer.json
wouterj: Yes, I'll move everything to PHP 8.5 and Symfony 8 once phpDocumentor supports it.

src/GuidesExtension/src/DocBuilder.php
wouterj: Yes, for now (to easily differentiate between new and old code), I've moved everything in the GuidesExtension. But some of it will be moved back once the repository is cleaned up.

composer.json
stof: but wouldn't this cause issues with the symfony.com website ? I doubt it still runs on Symfony 6.

wouterj: Yes, it runs the latest versions indeed. But only 1 small bugfix in phpDocumentor is required, I'm confident it supports it sooner than I'm ready for the symfony.com integration :)

composer.json
wouterj: Interesting, I thought we wanted to keep using simple-phpunit for libraries (especially by the Symfony team itself). But I see UX is using simple-phpunit, whereas AI is using phpunit. So we probably do not have a strict guideline on this

stof: We use `simple-phpunit` in the main mono-repo because of its built-in parallelization of tests per sub-package, that is quite specific to the structure of the main mono-repo.
For other projects, the `simple-phpunit` script does not provide much benefit anymore. It only makes it harder to control which version of PHPUnit is used and to ensure you don't have conflicting versions of dependencies installed (reduced risk as phpunit reduced its dependencies outside `phpunit/*` and `sebastian/*`, but not void as some other packages also reuse `sebastian/*`).
For this case, installing phpunit as a dev requirement is simpler.

src/GuidesExtension/src/Renderer/JsonRenderer.php
jaapio: @wouterj would a proper document tree in phpDocumentor/guides help?

wouterj: It would be nice for a fully compatible fjson renderer, but it's not a blocker for Symfony adaption as far as I can see (as we don't use the parent data).

wouterj:
Sooo, I've managed to hack this version of the DocsBuilder into the symfony.com repository. It's not nice yet, but it somewhat works.

As an example, let's take https://symfony.com/doc/8.1/quick_tour/the_architecture.html Locally with phpDocumentor Guides, this now renders like this: https://github.com/user-attachments/assets/da0d8b1f-7b5b-44ce-97df-4d5df05cd390

Comparing this with the live version, pixel by pixel, the only differences are found in the diff code blocks. And the phpDocumentor variant is actually more correct!

I've updated the JSON tests to represent the actual data we need: we need a lot of TOC information that I forgot to implement. Also, I want to refactor things a bit to merge `BuildEnvironment` and `BuildConfig`, and ideally avoid defining either as a service.

We also need better reporting on failures in `DocBuilder`, so we can make the build error/warning log working.

But, this is looking very promising!

src/GuidesExtension/src/Renderer/JsonRenderer.php
stof: As long as this renderer is in this repo (rather than being submitted upstream), it is not an issue if it does not support full compatibility with Sphinx (it might make sense to document the known differences though, for easier maintenance).

PR #204, opened 2026-04-14
Fix syntax highlighting of inline Twig comments

javiereguiluz:


stof:
I don't think this is the right fix. `# ` could also appear inside a string literal.
And Twig does not require inline comments to start with a space.

javiereguiluz:
You are right. I've proposed a different implementation.

stof (approved):

PR #206, opened 2026-07-11
Allow configuring the Symfony AI and UX versions used in class links

### LINKED ISSUE #199: Add support for versioning of Symfony AI and UX links
In #196 and #198 we added support for Symfony AI and UX links in `:class:` directives.

However, the generated URLs have the AI and UX versions hardcoded at `main` and `2.x` respectively.

This should be configuralbe, like the Symfony version is already configurable.

Amoifr:
The `:class:` directive supports Symfony AI (#196) and Symfony UX (#198) classes, but the generated URLs have their versions hardcoded to `main` and `2.x` in `ClassReference`. This makes them configurable, following the same pattern as the Symfony version.

`BuildConfig` gains two versions, defaulting to the current values, so nothing changes unless you opt in:

```php
$buildConfig = (new BuildConfig())
    ->setSymfonyVersion('7.4')
    ->setSymfonyAiVersion('1.0')   // defaults to 'main'
    ->setSymfonyUxVersion('3.x')   // defaults to '2.x'
;
```

The two URLs become placeholder constants (like `SYMFONY_REPOSITORY_URL`), `ClassReference` receives them instead of hardcoding the branches, and `KernelFactory` — its only instantiation site — passes them along.

About the tests:

* the existing `blocks/references/class` fixture passes untouched, which shows the defaults preserve the current output;
* the new tests cover both monorepos, including the slugged sub-repository case (`Symfony\AI\AiBundle` → `ai-bundle`);
* I double-checked they aren't vacuous: putting the branches back in `ClassReference` makes exactly the 3 configured AI/UX cases fail, while the Symfony one and the defaults stay green.

One thing I noticed while tracing the wiring, and deliberately left alone: `Application::__construct()` takes `$symfonyVersion` but never passes it to `BuildConfig`, so the `--symfony-version` CLI option looks like it has no effect. That's a separate matter — happy to open an issue if you'd like.

Fixes #199

Thanks for maintaining the tooling that keeps the docs building. 🙂

### DIFF
--- a/src/BuildConfig.php
+++ b/src/BuildConfig.php
@@ -18,10 +18,14 @@ class BuildConfig
     private const PHP_DOC_URL = 'https://secure.php.net/manual/en';
     private const SYMFONY_REPOSITORY_URL = 'https://github.com/symfony/symfony/blob/{symfonyVersion}/src';
     private const SYMFONY_DOC_URL = 'https://symfony.com/doc/{symfonyVersion}';
+    private const SYMFONY_AI_REPOSITORY_URL = 'https://github.com/symfony/ai/blob/{symfonyAiVersion}/src';
+    private const SYMFONY_UX_REPOSITORY_URL = 'https://github.com/symfony/ux/blob/{symfonyUxVersion}/src';
 
     private $useBuildCache;
     private $theme;
     private $symfonyVersion;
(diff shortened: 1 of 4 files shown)


javiereguiluz:
Thanks for your work on this new feature!

PR #208, opened 2026-07-11
Add a progress bar with a maximum for the rendering phase

### LINKED ISSUE #19: Fix the progress bar in build:docs command
The progress bar for the rendering process does not have a maximum - this should be fixed.

Amoifr:
Fixes #19, and removes the `// TODO: create a proper progress bar for rendering` in `BuildProgressListener`.

Today the parsing phase gets a real progress bar, while rendering only prints a static note:

```
 ! [NOTE] Rendering the HTML files...
```

Now it gets a bar that knows where it's going:

```
 ! [NOTE] Rendering 3 HTML files

 3/3 [============================] 100%
```

**How, since the parser has no per-document render event.** `Documents::render()` loops over the documents without dispatching anything, and `DocumentNode::renderDocument()` bypasses `Node::render()`, so there's no "document rendered" signal to hook into. But there is enough to work with:

* the **maximum** comes from `PreBuildRenderEvent` → `getBuilder()->getDocuments()->getAll()`, which is exactly the set of documents about to be rendered;
* the **progress** is tracked on `POST_NODE_RENDER` by looking at the file each rendered node belongs to, advancing once per file — the same "files already seen" approach the parsing bar already uses in `postParseDocument()`;
* the bar is finished on `POST_BUILD_RENDER`.

Nodes are also rendered outside the rendering phase, so the listener ignores them until the bar exists.

**Tests:** `BuildDocsCommandTest` now asserts both bars announce their file count and reach it. Verified the assertions fail without the change (the output only contains the old static note). Full suite green: 68 tests, 157 assertions.

---

Unrelated, but I ran into it while testing this and it seems worth flagging: **`bin/docs-builder` is currently broken** on Symfony 8 — `Application::add()` was removed in favour of `addCommand()`, so the binary dies immediately:

```
PHP Fatal error: Call to undefined method Symfony\Component\Console\Application::add() in src/Application.php:38
```

The test suite doesn't catch it because it drives the command through `CommandTester` rather than the binary. Happy to send a separate PR if you'd like.

### DIFF
--- a/src/Listener/BuildProgressListener.php
+++ b/src/Listener/BuildProgressListener.php
@@ -12,6 +12,8 @@ declare(strict_types=1);
 namespace SymfonyDocsBuilder\Listener;
 
 use Doctrine\Common\EventManager;
+use Doctrine\RST\Event\PostBuildRenderEvent;
+use Doctrine\RST\Event\PostNodeRenderEvent;
 use Doctrine\RST\Event\PostParseDocumentEvent;
 use Doctrine\RST\Event\PreBuildParseEvent;
 use Doctrine\RST\Event\PreBuildRenderEvent;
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Thank you Pascal.

